Issue #31, opened 2026-05-19
[Feature]: --dry-run on gwm remove and gwm prune


kbrdn1:
## Problem

\`gwm prune\` walks the worktree list, detects prunables, and removes the admin entries — all in one shot, no preview. \`gwm remove auth\` resolves a fuzzy pattern then destroys, no chance to verify the resolution matched what the user meant. Both commands deserve a \"show me what you would do\" mode.

## Proposed solution

Add a \`--dry-run\` flag to both:

- \`gwm prune --dry-run\` lists every prunable worktree (name + path + reason) and exits without touching anything. Exit code 0.
- \`gwm remove <pattern> --dry-run\` prints the resolved worktree (name + path + branch + would-delete-branch flag), then exits. If the pattern is ambiguous, exits non-zero with the candidate list — same as the regular failure mode.

The output is plain-text aligned, so it can be piped into a confirmation script for users who want belt-and-suspenders.

## Alternatives considered

- A separate \`gwm preview prune\` sub-command: more verbose for the user, more code paths to maintain.
- Always prompt interactively before destruction: noisy for users who know what they're doing; \`--dry-run\` keeps confirmation opt-in.

## Surface

cli (new / changed subcommand)

## Backward compatibility

No breaking change.

## Example usage

\`\`\`bash
gwm prune --dry-run
would prune 2 worktree(s):
  feat-old-1   /Users/x/cc-worktree/myrepo/feat-old-1   (working dir missing)
  fix-leftover /Users/x/cc-worktree/myrepo/fix-leftover (working dir missing)

gwm remove auth --delete-branch --dry-run
would remove:
  name:   feat-99-user-authentication
  path:   /Users/x/cc-worktree/myrepo/feat-99-user-authentication
  branch: feat/#99-user-authentication (would be deleted)
\`\`\`

## Additional context

Implementation is a thin guard in the existing \`worktree::remove\` and \`worktree::prune\` paths: gather the work, print the plan, return early. Same surface, opt-in flag.

kbrdn1:
[...]

PR #155, opened 2026-05-23
✨ feat(cli): gwm undo + gwm history + operation journal (#29)

### LINKED ISSUE #29: [Feature]: gwm undo — recover the last destructive operation
## Problem

\`gwm remove\` and (worse) \`gwm remove --delete-branch\` are easy to misfire: a stray \`d\` keypress in the TUI confirm dialog with \`p\` already toggled, or a typo on the CLI fuzzy match resolving to the wrong worktree. Today the only recovery is \`git reflog\` + manual \`git branch\` recreation + \`gwm create\` again — error-prone and slow.

## Proposed solution

Maintain a tiny operation journal in \`\$XDG_DATA_HOME/gwm/history.toml\` (or \`~/.local/share/gwm/history.toml\`). Each destructive operation appends an entry:

\`\`\`toml
[[op]]
ts = \"2026-05-19T08:42:11Z\"
kind = \"remove\"
worktree = \"feat-99-user-authentication\"
branch = \"feat/#99-user-authentication\"
branch_oid = \"a1b2c3d4...\"
path = \"/Users/x/cc-worktree/myrepo/feat-99-user-authentication\"
deleted_branch = true
\`\`\`

\`gwm undo\` reads the last entry, re-creates the branch at \`branch_oid\`, re-adds the worktree at \`path\`, and (if the journal has the bootstrap report fingerprint) optionally re-runs bootstrap. Removes the entry on success.

\`gwm history\` lists the last N operations, with their timestamps and a one-line summary.

Capped at, say, 100 entries with a TOML-aware rotation. Per-repo separation (the journal includes the repo path so undo doesn't cross-pollute).

## Alternatives considered

- A trash directory (move the worktree to \`~/.local/share/gwm/trash/\` instead of deleting): heavier on disk, weird for users who clean their home dir aggressively. The journal records the SHA; \`git\` already keeps the commit objects until gc.
- Rely on \`git reflog\`: doesn't recover the worktree admin entry under \`.git/worktrees/\`; user still has to re-run \`git worktree add\`.

## Surface

cli (new / changed subcommand)

## Backward compatibility

No breaking change.

## Example usage

\`\`\`bash
gwm remove auth --delete-branch
# oh no.

gwm undo
✓ recreated branch feat/#99-user-authentication at a1b2c3d4
✓ re-added worktree at /Users/x/cc-worktree/myrepo/feat-99-user-authentication
(skipped re-bootstrap; pass --bootstrap to run it)

gwm history
1m ago  remove   feat-99-user-authentication  (undone)
2h ago  remove   chore-12-ci-dev-prerelease
5h ago  create   feat-10-tui-lazygit-sidebar
\`\`\`

## Additional context

Branch resurrection uses \`Repository::reference\` to point at the saved OID (still in the object DB unless a \`git gc\` happened). Worktree resurrection: \`Repository::worktree\` against the saved path. Bootstrap re-run is opt-in to keep undo cheap.

kbrdn1:
## Description

Implements [issue #29](https://github.com/kbrdn1/gwm-cli/issues/29):
a tiny operation journal that lets `gwm undo` recover from a stray
`gwm remove` (with or without `--delete-branch`) without `git
reflog` archaeology, plus `gwm history` for inspection.

Closes #29

## Type of change

## Changes

- New `src/history.rs` module: `Journal` (load / append / save /
  rotate at 100 entries) + `OpEntry` + `default_journal_path()`
  (resolution order: `$GWM_HISTORY_FILE` → `$XDG_DATA_HOME/gwm/
  history.toml` → `dirs::data_dir()/gwm/...`). Atomic save via
  the same tempfile-rename primitive `trust::TrustLedger::save`
  uses.
- New `gwm history [--limit N] [--all]` subcommand. Default
  `--limit` is 20; filters to the current repo's canonicalised
  workdir by default, `--all` opts out for forensic /
  multi-repo grep. Empty result prints `no operations
  recorded` (stable scripted signal). Newest-first sort.
- New `gwm undo [--bootstrap]` subcommand. Pops the most
  recent op for the current repo, recreates `refs/heads/
  <branch>` at the saved OID via `Repository::reference`,
  re-adds the worktree at the saved path with `reuse_branch:
  true`. The `--bootstrap` flag opts into re-running the
  per-worktree bootstrap (off by default — undo stays
  cheap). The journal entry is consumed only AFTER a
  successful resurrection so a mid-flight failure keeps the
  recovery anchor intact.
- Hook in `cmd_remove`: captures `branch_oid` via libgit2
  BEFORE the destructive call, writes the journal entry,
  then runs the destruction. Fires only on the destructive
  path — `--dry-run` short-circuits before reaching the
  hook so a preview cannot pollute the journal. Journal IO
  failures are logged to stderr but NEVER block the remove.
- `help_prints_subcommands` updated to assert both new
  subcommands appear in `gwm --help` (the canary contract
  per `CLAUDE.md`).
- `CHANGELOG.md` entry under `[Unreleased] / Added`.

## Tests

- [x] `cargo test` passes locally (838 passed, 0 failed)
- [x] New tests added under `tests/`:
  - `tests/history_tests.rs`: 8 unit tests pinning the
    journal IO contract (round-trip, rotation cap, per-repo
    filter, last/pop helpers, env-based path resolution).
  - `tests/cli_binary.rs`: 8 new E2E tests pinning the
    `gwm history`, `gwm undo`, and `gwm remove`-hook
    contracts (empty journal, single-op listing, per-repo
    filter, `--all` opt-out, undo round-trip, undo on empty
    journal, hook on destructive remove, no-hook on
    dry-run).
- Pre-validated under a stripped PATH (`PATH=\"\$(dirname \$(command -v cargo)\):/usr/bin:/bin\"`)
  per `CLAUDE.md` — every env-sensitive test isolates via
  `GWM_HISTORY_FILE` and a `tempfile::TempDir`.

## Design decisions

- **Rotation is global, not per-repo.** Keeps the
  bookkeeping trivial (one cap to enforce on every
  append). A heavy user juggling 30 repos still gets a
  bounded ledger; the per-repo separation is enforced at
  READ time via `entries_for_repo`, not at write time.
- **Hook only on `remove` today.** The `OpKind` enum
  reserves a `Create` variant for a future extension that
  would let `gwm undo` also roll back accidental worktree
  creations, but wiring it in this PR would mean two more
  destructive surfaces to design (does undo of \`create\`
  delete the dir? what about uncommitted work?). Out of
  scope for #29; can land as a follow-up if real usage
  justifies it.
- **`--bootstrap` opt-in on undo.** Per the issue's design
  notes: bootstrap can be expensive and the user often
  just wants the directory back. The opt-in keeps the
  default path fast and predictable.
- **Atomic resurrect-then-persist.** `Journal::save` runs
  only after both branch + worktree are back. A mid-flight
  failure leaves the entry intact so the user can retry
  (idempotent against the "branch already exists" path).
- **`eprintln!` on journal IO failure, never block.** The
  destruction is what the user typed; losing recoverability
  is unfortunate, blocking the remove because we couldn't
  write `~/.local/share/gwm/history.toml` (disk full,
  read-only FS, sandboxed CI runner) would be far more
  surprising.

## Smoke verification

```bash
$ gwm history
no operations recorded

$ gwm undo
error: nothing to undo for /Users/kbrdn1/Projects/Perso/gwm-cli — the journal is empty for this repo

$ gwm --help
...
  history        List the recent destructive operations recorded by `gwm` (issue #29)...
  undo           Undo the most recent destructive operation recorded for the current repo...
```

Full create → remove → undo round-trip is covered by
`undo_recreates_branch_and_worktree_after_remove` in
`tests/cli_binary.rs`.

## Checklist

- [x] No `unwrap()` on user-facing paths (only in test
      helpers / genuinely-infallible call sites)
- [x] No `println!` in TUI render code (journal writes
      happen at the CLI layer; `worktree::remove` is
      untouched and the warning goes through `eprintln!`
      at the CLI side)

## Linked issues / docs

- Issue: #29
- Related: #31 (`--dry-run` flag whose gate this PR honours)

## Notes for reviewers

- The TDD pair pattern is strict: every commit alternates
  RED (test pinning behaviour, builds-but-fails) and
  GREEN (minimal production code to flip the test green).
  The 8 commits are atomic and reversible.
- Per-repo separation is enforced via canonicalised
  `repo_root` paths — important on macOS where `/var` and
  `/private/var` are the same inode but compare unequal as
  strings.
- Follow-up ideas (filed only if reviewers see real
  scope): hooking `gwm create` into the journal so undo
  can also roll back accidental creations; recording a
  `bootstrap_fingerprint` so `--bootstrap` knows whether
  the prior bootstrap report would still apply.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,44 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`gwm undo` + `gwm history` + operation journal**
+  ([#29](https://github.com/kbrdn1/gwm-cli/issues/29)). Recover from
+  a misfired `gwm remove` without `git reflog` archaeology.
+  - **Operation journal**: every destructive op (`gwm remove`,
+    with or without `--delete-branch`) appends an entry to
+    `$XDG_DATA_HOME/gwm/history.toml` (override via
+    `$GWM_HISTORY_FILE`; macOS fallback under `Application
+    Support`, Windows under `%LOCALAPPDATA%`). Each entry
+    captures the timestamp, worktree name, branch name,
+    branch OID at deletion time, on-disk path, the
+    `--delete-branch` flag, and the canonicalised
+    `repo_root` so per-repo separation works across symlink
+    chains.
+  - **Rotation cap**: 100 entries max across the whole
+    journal; oldest-by-timestamp dropped on overflow.
+  - **`gwm undo [--bootstrap]`**: pops the most recent op for
+    the current repo, recreates `refs/heads/<branch>` at the
+    saved OID via `Repository::reference`, re-adds the
+    worktree at the saved path with `reuse_branch: true`. The
+    `--bootstrap` flag opts into re-running the per-worktree
+    bootstrap (off by default to keep undo cheap). The
+    journal entry is consumed only after a successful
+    resurrection — a mid-flight failure keeps the recovery
+    anchor intact.
+  - **`gwm history [--limit N] [--all]`**: lists the recent
+    ops newest-first. Default `--limit` is 20. By default
+    filters to the current repo's canonicalised workdir;
+    `--all` surfaces every entry across every repo for
+    forensic / multi-repo grep. Empty result prints `no
+    operations recorded` as a stable scripted signal.
+  - `gwm remove --dry-run` does NOT write to the journal —
+    previewing a destruction must never allow "undoing"
+    something that never happened.
+  - Journal IO failures are logged to stderr but never block
+    the destructive op: failing a remove because we couldn't
+    write `~/.local/share/gwm/history.toml` (disk full,
+    read-only FS, sandboxed CI runner) would be more
+    surprising than losing recoverability.
 - **`--dry-run` flag on `gwm remove` and `gwm prune`**
   ([#31](https://github.com/kbrdn1/gwm-cli/issues/31)). Preview
   destructive operations before running them.
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -834,7 +834,9 @@ fn cmd_remove(pattern: String, delete_branch: bool, dry_run: bool) -> Result<()>
     // already happened above — an ambiguous pattern surfaced via
     // `find_fuzzy` returns the same `Other(... ambiguous ...)` error
     // the destructive form raises, satisfying the spec's "same error
-    // contract" requirement.
+    // contract" requirement. The journal hook MUST NOT fire here —
+    // a preview that wrote to the journal would let the user "undo"
+    // something that never happened.
     worktree::remove_dry_run(&repo, &found.name)?;
     print!(
       "{}",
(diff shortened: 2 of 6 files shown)


kbrdn1:
## Update — Copilot review applied

All four Copilot comments resolved in commits \`97b09f7\` and \`967e8ae\`:

**Already fixed before the Copilot review landed** (\`97b09f7\`):

- The two \`tests/cli_binary.rs\` "TOML escaping on Windows" comments — \`repo_root\` is now routed through \`toml_basic_string()\` at every test seed site. Confirmed green on \`test (windows-latest)\`.

**Fixed in \`967e8ae\`**:

- \`tests/history_tests.rs:245\` env-mutation safety — adopted the \`env_lock()\` pattern from \`tests/trust_tests.rs\` (process-global \`Mutex\` + \`unsafe\` blocks around \`set_var\` / \`remove_var\`), so the test is forward-compatible with the Rust 1.86+ \`unsafe-env-mutation\` rule.
- \`src/cli.rs:2208\` — \`gwm history --limit 0\` no longer falsely prints "no operations recorded" when ops exist. The two cases are now disambiguated: empty journal prints the sentinel, \`--limit 0\` on a non-empty journal prints nothing and exits 0.
- \`src/cli.rs:2271\` — \`gwm undo\` no longer falls back to \`branch_name = "HEAD"\` for detached-HEAD entries. It surfaces a clear error naming the worktree and the limitation; detached-HEAD resurrection is a follow-up if real usage justifies it.

Two regression guards added: \`history_with_zero_limit_prints_nothing_but_does_not_lie\` and \`undo_refuses_detached_head_entry_with_clear_error\`.

CI matrix is fully green: clippy, rustfmt, ubuntu-latest, macos-latest, windows-latest, pre-commit hook smoke, cargo audit, gwm doctor (advisory), GitGuardian. Total test count: 840 passed, 0 failed.

Issue #258, opened 2026-06-07
[Feature]: Expose `gwm sync` as a TUI action


kbrdn1:
## Problem

`gwm sync` (issue #24) exists **only as a CLI subcommand** (`src/sync.rs`). There is no TUI surface for it: no `Action::Sync` in `src/tui/keymap.rs`, no keybinding, no `Action::Sync` handler. To rebase/merge a worktree's branch onto its upstream while inside the TUI, the user has to quit (or shell out) and run `gwm sync` by hand.

This also surfaced as a scoping gap for #231 (async-task layer): the original `/goal` text asked for `refresh + sync + bootstrap + launcher` to all become non-blocking through the shared loader. `refresh` and (follow-ups) `bootstrap`/`github` map onto the spine cleanly, but **`sync` has no TUI action to make non-blocking in the first place** — making it "non-blocking" presupposes first building the TUI action. That work is deliberately carved out here as a separate feature rather than smuggled into #231.

## Proposed solution

Add a first-class `sync` action to the TUI:

1. **Keymap** — new `Sync => \"sync\"` variant in `define_actions!` (`src/tui/keymap.rs`), default binding `s` (currently unbound), rebindable via `[tui.keys]`.
2. **Handler** — on the selected worktree, run the existing `sync::sync_worktree` logic (read-side libgit2 inspection + shell-out for fetch/rebase/merge), reporting outcome (`AlreadyLevel` / `FastForwarded` / `Rebased` / `Merged` / dirty-refused) through the status bar — the only runtime feedback channel in the TUI.
3. **Strategy** — default rebase (repo convention); a `--merge` equivalent can come later (e.g. via the open-menu) and is out of scope here.
4. **Async** — run it off-thread on the #231 `TaskRunner` spine (new `TaskKind::Sync`) so it does not block the event loop, rendering through the shared statusbar spinner like refresh. (This is what closes the original `/goal` gap.)
5. **Post-sync refresh** — bump the refresh generation on completion so the worktree list/status reflects the new ahead/behind state (same race guard as `App::refresh()`).

## Alternatives considered

- **Route it through the open-menu (`o`) instead of a dedicated key** — viable, but `sync` is a frequent lifecycle verb on par with `refresh`/`bootstrap`, which both have top-level keys. A dedicated rebindable action is the consistent choice; the open-menu can still offer a `sync --merge` variant later.
[...]

Issue #248, opened 2026-06-06
[Bug]: flaky `refresh_github_status_auto_detects_pr_for_unlinked_branch` on CI (env/parallel race)


kbrdn1:
## Description

`tests/tui_app_tests.rs::refresh_github_status_auto_detects_pr_for_unlinked_branch` intermittently fails on CI (observed on `test (ubuntu-latest)` for PR #246, while the same test passed on the ubuntu runs of #245 and #247 against the same `dev` base):

```
thread '...auto_detects_pr_for_unlinked_branch' panicked at tests/tui_app_tests.rs:1813:
assertion `left == right` failed: a detected PR must re-resolve on refresh
  left: None
  right: Some(200)
```

## Root cause (suspected)

The test writes a fake `gh` script to a TempDir and points `GWM_GH` (a **process-global** env var) at it, guarded by `env_lock()`. The auto-detect path (`github::find_pr_for_branch`) is synchronous, so this is **not** an off-thread race. The likely cause is a **parallel-test env/cwd race**: cargo runs tests in the same binary on multiple threads, and a *different* test that mutates `PATH`/cwd/env **without** taking `env_lock()` can transiently break resolution of the fake `gh`, yielding `None`. Under ubuntu-runner load the window widens, hence the intermittent red.

This is **not** a regression from #246 (RepoContext, `src/cli.rs`) — that PR does not touch the TUI GitHub-refresh path.

## Acceptance criteria

- [ ] Make the test robust to parallel env mutation: either gate the fake-`gh` resolution on something not shared process-wide, ensure **every** env/cwd-mutating test in the binary takes the same `env_lock()`, or assert intent without depending on global `GWM_GH` (per CLAUDE.md's env-independent-test rule).
- [ ] `cargo test` green across 3 consecutive CI runs (no skip / `#[ignore]` — fix the race, not the symptom).

## Notes

Surfaced during the refactor-batch-1 CI pass. Re-running the failed job unblocks #246's merge in the meantime.

kbrdn1:
Fixed via #265 (merged into dev). Root cause: off-thread `pr view` worker still execing the shared fake-gh script when the test rewrote+reexec'd it (Linux ETXTBSY), not the suspected env-var race. Fixed with distinct write-once scripts.

Issue #231, opened 2026-06-06
[Feature]: async-task layer for slow ops + better loader render


kbrdn1:
## Problem

#217 moved the GitHub issue/PR fetch (`F`) off-thread so the statusbar spinner animates instead of the event loop blocking — but that was a single call site. Other potentially-slow operations still run **synchronously on the event loop**: the worktree list refresh (`f`/`r`), `gwm sync`, bootstrap re-runs (`b`), and the launcher spawns. When those run on a large repo or a slow network, the TUI stalls (no frame, no spinner, input queues). The loader itself is also minimal — a single statusbar spinner with no per-operation context, no progress, and inconsistent rendering across the places that do show "loading".

## Proposed solution

Generalise the off-thread fetch pattern into a **shared async-task + loader layer**, and improve the loader render.

**Async threading**
- A small task runner (mirroring `state/github_fetch.rs`) so any slow op — list refresh, `sync`, bootstrap, launcher prep — runs on a worker thread and posts its result back to the event loop. The loop keeps rendering (spinner animates, `q`/`Esc` stay responsive); late results are dropped on invalidation, like the existing `GitHubFetch` cache.
- Clear in-flight / done / failed states per task, surfaced in the statusbar context chip.

**Better loader render**
- A reusable loader widget (consistent spinner glyphs + label) used by every async site instead of ad-hoc strings, so "loading" looks the same everywhere.
- Per-operation label ("fetching GitHub…", "refreshing worktrees…", "syncing…") and, where the underlying op exposes it, coarse progress; graceful failed-state render.
- Builds on `state/spinner.rs`; no `println!` in render (statusbar-only rule).

## Alternatives considered

- **Leave each call site synchronous**: simplest, but every slow op freezes the UI — the exact problem #217 fixed for one case.
- **Full async runtime (tokio)**: overkill for a handful of shell-outs / libgit2 calls; a worker-thread + channel keeps the dependency surface and complexity low.

## Surface

tui (new view, keybinding, behaviour)

## Backward compatibility

No breaking change — internal threading + render; same keys, same outputs, just non-blocking with a consistent loader.

## Example usage

```
[TUI, large repo / slow network]
<f>      — list refresh runs off-thread; spinner animates, statusbar shows "refreshing worktrees…"
<F>      — GitHub fetch (already off-thread) now renders through the same shared loader
<q>      — stays responsive while any task is in flight
```

## Additional context

[...]

Issue #95, opened 2026-05-22
[Feature]: TOFU trust ledger on .gwm.toml + --allow-bootstrap flag


kbrdn1:
## Problem

`gwm` reads `.gwm.toml` from the checked-out repository and runs `[[bootstrap.command]]` shell lines under the user's privileges — no prompt, no allowlist, no signing. Cloning an arbitrary repository and running `gwm create` (or any command that triggers bootstrap) is equivalent to `curl … | sh` against the repo author. There is no trust boundary between "I trust this remote" and "this remote can execute commands as me".

The threat is not hypothetical: every CI runner that does `git clone … && gwm create …` to spin up an ephemeral workspace is exposed; every contributor cloning a PR fork inherits the fork's bootstrap commands without ever being prompted; every team using `gwm` on a shared monorepo trusts every co-worker with commit access on `.gwm.toml`.

## Proposed solution

A TOFU (trust-on-first-use) trust ledger at `~/.config/gwm/trust.toml` keyed on (origin URL, hash of `.gwm.toml`), plus a global `--allow-bootstrap` / `--deny-bootstrap` flag.

```toml
# ~/.config/gwm/trust.toml — managed by gwm, hand-edit at your own risk
[[entries]]
origin       = "git@github.com:kbrdn1/gwm-cli.git"
config_sha   = "3a4f9c…"
trusted_at   = "2026-05-22T10:00:00Z"
trusted_by   = "kylian@laptop"
```

Behaviour:

1. **First run on a repo:** parse `.gwm.toml`, hash it, look up `(origin, hash)` in the ledger. If missing, print a summary of what bootstrap will execute (copy targets, guard patterns, shell command lines) and prompt `Trust this .gwm.toml? [y/N/show]`. `show` re-prints the toml; `y` records the entry and proceeds; `N` aborts.
2. **Subsequent runs:** silent pass — the hash matches.
3. **Hash drift** (file edited since last trust): re-prompt with a unified diff of the change.
4. **CI bypass:** `--allow-bootstrap` skips the prompt (off by default, opt-in). `GWM_ALLOW_BOOTSTRAP=1` env equivalent for non-interactive runners.
5. **`gwm trust list / revoke / show`** subcommands to manage the ledger.

The hash is over the raw bytes of `.gwm.toml`, so even whitespace changes re-prompt — intentional, since whitespace in a shell command line can change semantics (`rm -rf /tmp/` vs `rm -rf /tmp /`).

## Alternatives considered

[...]

PR #266, opened 2026-06-07
✨ feat(tui): run bootstrap (`b`) off-thread on the async-task spine

### LINKED ISSUE #256: [Task]: move bootstrap (b) off-thread via the async-task spine (#231 follow-up)
Follow-up to #231 (merged in #254). `App::bootstrap_selected` runs `bootstrap::run` synchronously on the event loop and then transitions to `View::Report`. On a slow bootstrap (command hooks, copies) this blocks the TUI.

**Goal:** run bootstrap on the `TaskRunner` spine like the worktree refresh.

**Notes**
- Add a `TaskKind::Bootstrap` + a `TaskMsg` variant carrying `Result<BootstrapReport, String>`.
- The completion is more invasive than refresh: it sets `self.report` and flips to `View::Report` — decide what the UI shows *while* it runs (statusbar loader is in place; the Report view appears on drain).
- Worker must move only owned `Send` data (clone `workdir`/`path`/`config`); `bootstrap::run` borrows a `BootstrapCtx`.
- TOFU trust gate stays on the main thread before spawning.

kbrdn1:
## Description

Follow-up to #231. `bootstrap_selected` (the `b` key — re-run bootstrap on the selected worktree) ran `bootstrap::run` (file copies, guards, command hooks) **synchronously on the event loop**, freezing the TUI on a slow bootstrap. This moves it onto the shared `TaskRunner` spine, exactly like the worktree refresh (#231) and sync (#258).

Closes #256

## Type of change

- [x] ✨ Feature (non-blocking bootstrap; new `TaskKind`)

## Changes

- `src/tui/state/async_task.rs`: `TaskKind::Bootstrap` (single global op → coalesces; `bootstrapping…` label) + `TaskMsg::Bootstrap(generation, Result<BootstrapReport, String>)`.
- `src/tui/app.rs`: `bootstrap_selected` keeps the synchronous TOFU trust gate on the main thread, then claims a generation and spawns `spawn_bootstrap` (moves only owned `Send` data — the two paths + a `Config` clone — and rebuilds its own `BootstrapCtx`). The `View::Report` transition + ok/had-failures status move to `drain_task_results`, behind the generation late-drop guard.
- Tests + docs (EN/FR keybindings, CHANGELOG).

## Tests

- [x] `cargo test` passes locally (env-minimal PATH; 60 binaries green)
- [x] New tests added (TDD red→green): spine contract (label / coalesce / late-drop) in `tui_state_async_task_tests.rs`; drain behaviour (Ok→Report, Failed→"had failures", Err→status no-flip, late-drop, always-flip-even-mid-create, no-selection guard) in `tui_app_tests.rs`. No real OS thread spawned in tests (results injected via `task_result_sender` + drain — the #248 trap avoided).

## Notes for reviewers

- **Intentional scope:** only the `b` re-run (`bootstrap_selected`) moves off-thread. The **create flow** (`n` → `submit_create`) still calls `bootstrap::run` synchronously — the issue names `bootstrap_selected` specifically, and migrating the create flow (which interleaves with the create modal's own state machine) is a separate, differently-sequenced change. Possible follow-up; deliberately **not** in this PR.
- **Behaviour change vs the old sync path:** because the run is now async, the user can navigate between the `b` press and the result. A live result **always flips to `View::Report`** — the user asked for the bootstrap, so its outcome takes the screen even if they opened the create form meanwhile (the create slice persists; only the modal context is lost on the next `Esc`). This is pinned by `drain_bootstrap_report_flips_to_report_even_from_another_view`. If we'd rather protect an open modal, the minimal guard (flip only from `View::List`) is a one-liner follow-up.
- TOFU trust gate unchanged (#95): still synchronous, still before the spawn — the existing `tui_bootstrap_selected_aborts_on_untrusted_config` stays green.

## Linked issues / docs

- Issue: #256
- Built on: #231 (async-task spine), mirrors #258 (sync)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-_No changes yet — entries land here as PRs merge into `dev`, then move to a per-RC file under `changelogs/pre-releases/` when the next RC is cut._
+### Changed
+
+- Off-thread bootstrap: pressing `b` to re-run bootstrap on the selected
+  worktree now runs `bootstrap::run` (file copies, guards, command hooks)
+  on the shared async-task spine instead of blocking the event loop, so a
+  slow bootstrap no longer freezes the TUI. The statusbar spinner animates
+  with a `bootstrapping…` label and `q` / `Esc` stay responsive while it
+  runs; the Report view opens when it finishes. The TOFU trust gate still
+  runs synchronously on the main thread before the worker spawns, and a
+  second `b` press while one is in flight coalesces (no double run).
+  ([#256](https://github.com/kbrdn1/gwm-cli/issues/256))
 
 ## Past releases
 
--- a/docs/2.tui/1.keybindings.md
+++ b/docs/2.tui/1.keybindings.md
@@ -25,7 +25,7 @@ The full key map for `gwm`'s ratatui interface. Press `?` at any time for the sa
 | `G`         | jump to the last worktree                                                                         |
 | `n`         | new worktree (form: type → issue → description) · gated by the [TOFU trust ledger](/configuration/trust-ledger) — refuses with a status-bar hint on untrusted `.gwm.toml` |
 | `d`         | delete selected (confirm `y` · countdown when `p` is armed — see [confirm-overlay](/tui/confirm-countdown)) |
-| `b`         | re-run bootstrap on the selected worktree · same [trust gate](/configuration/trust-ledger) as `n` |
+| `b`         | re-run bootstrap on the selected worktree, off-thread (the statusbar spinner animates while it runs; the Report view opens when it finishes) · same [trust gate](/configuration/trust-ledger) as `n` |
 | `o`         | open the worktree per [`[tui.open]`](/tui/open-dispatch) — `shell` (default) / `editor` / `finder` |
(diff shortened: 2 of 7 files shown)


kbrdn1:
### Codex review — P1 acknowledged, consciously deferred to #267

Codex flagged a [P1] on `spawn_bootstrap`: a detached worker means quitting (`q`/`Esc`) mid-bootstrap abandons the run, leaving a partially-applied worktree. Valid finding.

**Deferred to #267** (not fixed here), because the *correct* fix is generic, not bootstrap-specific:

- The same pattern already ships in **`spawn_sync` (#258, merged)** — `sync::sync` (fetch + rebase) is equally mutating and equally abandonable. So this is a pre-existing **class** of issue, not one #256 introduces.
- The right-altitude fix (`TaskKind::is_mutating()` + gate the quit path on in-flight mutating tasks) changes the runtime behaviour of already-merged sync — a cross-cutting run-loop lifecycle change that does **not** belong buried in a feature PR titled "bootstrap off-thread". A bootstrap-only guard would be exactly the special-case this PR otherwise avoids.
- Severity isn't 🔴: bootstrap copies files into a freshly-created worktree and runs re-runnable setup hooks — worst case is a partial copy (overwritten on the next `b`) plus a possibly-orphaned hook. It never touches the user's source tree and leaves nothing unrecoverable.

Tracked in **#267** (scoped to both `Bootstrap` and `Sync`, with its own quit-while-mutating-waits / quit-while-read-only-doesn't tests). Merging #256 atomic on green.

PR #268, opened 2026-06-07
✨ feat(tui): open the gwm docs in the browser with `.`

### LINKED ISSUE #233: [Feature]: open the docs in the browser, bound to `.`
## Problem

gwm ships a full docs site (`docs/`, EN + FR) and a public repo, but from inside the TUI there is no quick way to reach them. A user who hits an unfamiliar key or wants the config reference has to leave the terminal and find the URL manually. The TUI already opens external targets in the browser (`O` for the linked issue/PR), so an "open the docs" shortcut is a small, natural addition.

## Proposed solution

A **`.` keybinding that opens the gwm documentation in the browser** (the docs site; fall back to the repo's `docs/` on GitHub). Mirrors the existing `O` open-in-browser plumbing.

- `.` opens the docs landing page in the default browser (same spawn path as `O` / `gwm open`).
- Rebindable action (e.g. `open_docs`) under `[tui.keys]`.
- Optionally context-aware later (open the keybindings page from the help overlay, the themes page from theme mode, …) — MVP is just the landing page.
- A `[docs]` config knob (or reuse of the repo URL) can override the target for forks; MVP can hard-default to the canonical docs URL.

## Alternatives considered

- **Only document the URL in the help overlay**: no action, still a manual copy-paste. A keybinding is one keystroke.
- **Render docs inside the TUI**: out of scope — docs are rich Markdown + assets; the browser is the right surface.

## Surface

tui (new view, keybinding, behaviour)

## Backward compatibility

No breaking change — additive key (`.`), reuses the existing browser-spawn path.

## Example usage

```
[TUI]
<.>      — opens the gwm docs in the default browser
```

## Additional context

Smallest of the near-term TUI items and dependency-free — reuses the `O` / `gwm open` browser-spawn helper. Complements the in-TUI Configuration panel (`4`): docs for the prose reference, the panel for the live resolved config.

kbrdn1:
## Description

Adds a `.` keybinding that opens the gwm documentation in the default browser, reusing the existing OS-opener path (the one behind the `O` issue/PR open menu). The smallest of the near-term TUI items, dependency-free.

Closes #233

## Type of change

- [x] ✨ Feature (additive key `.` + new `open_docs` action)

## Changes

- `src/tui/keymap.rs`: `Action::OpenDocs`, default binding `.` (rebindable as `open_docs`).
- `src/tui/mod.rs`: `pub const DOCS_URL = concat!(env!("CARGO_PKG_REPOSITORY"), "/tree/main/docs")` + dispatch `Action::OpenDocs` → `open_url(DOCS_URL, app)` (gated `!picker_mode`).
- `src/tui/palette.rs`: `:open-docs` registry entry (also required by the `registry_covers_every_action_variant` parity test).
- `src/tui/ui.rs`: help-overlay row.
- Tests + docs (EN/FR keybindings, CHANGELOG).

## Tests

- [x] `cargo test` passes locally (env-minimal PATH; 60 binaries green)
- [x] New tests: `.`→`OpenDocs` default + rebindable (keymap_tests); `:open-docs` entry + `DOCS_URL` canonical shape (palette_tests). The change is additive; behaviour is pinned by these + the existing palette parity test (which fails if the action isn't wired through the registry).

## Notes for reviewers

- **No docs site deployed**, so MVP targets the repo's `docs/` tree on GitHub (`/tree/main/docs`). Derived from `CARGO_PKG_REPOSITORY` so a fork resolves to its own docs without a patch. A `[docs]` config override (issue's optional idea) is a possible follow-up, not in MVP.
- `.` only triggers `OpenDocs` in the normal list view — inside the filter bar it stays filter input (handled before the action dispatch), and it's gated `!picker_mode` like the other list-view actions.
- Context-aware targets (open the keybindings page from Help, themes page from theme mode) are explicitly deferred per the issue — MVP is the landing page.

## Linked issues / docs

- Issue: #233

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Open docs in the browser: press `.` in the TUI to open the gwm
+  documentation in your default browser, reusing the same OS-opener path
+  as the `O` issue/PR open menu. The target is the repo's `docs/` tree on
+  GitHub, derived from the crate's `repository` so a fork resolves to its
+  own docs; rebindable via `[tui.keys] open_docs` and reachable by name
+  through `:open-docs`. ([#233](https://github.com/kbrdn1/gwm-cli/issues/233))
+
 ### Changed
 
 - Off-thread bootstrap: pressing `b` to re-run bootstrap on the selected
--- a/docs/2.tui/1.keybindings.md
+++ b/docs/2.tui/1.keybindings.md
@@ -31,6 +31,7 @@ The full key map for `gwm`'s ratatui interface. Press `?` at any time for the sa
 | `l`         | launch the configured [`[git_tui]`](/tui/launchers) command (default `lazygit -p {path}`, fullscreen) |
 | `R`         | launch the configured [`[review]`](/tui/launchers) command — AI / web reviewer over `git diff base..head` |
 | `O`         | open menu for the linked issue / PR (`i` issue, `p` pr → spawns browser)                          |
+| `.`         | open the gwm documentation in the default browser (rebindable action `open_docs`)                 |
 | `L`         | link prompt — type `i` or `p`, then digits, to attach an issue / PR                               |
 | `F`         | refresh the GitHub issue / PR status (off-thread `gh` fetch — the statusbar spinner animates while it runs) |
 | `f`         | refresh the worktree list (alias: `r`)                                                            |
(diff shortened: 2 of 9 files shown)


kbrdn1:
### Codex review — P2 fixed

Codex flagged that `OpenDocs` was gated on `!picker_mode`, so `.` / `:open-docs` would silently no-op in `gwm switch` even though the help overlay advertises it. Correct — opening docs is **read-only and selection-independent**, exactly like `open` / `yank` / `git_tui` (none of which are picker-gated).

Fixed by removing the picker-mode gate (folded into the feat commit; branch force-pushed). `.` now opens the docs in picker mode too, consistent with the other read-only actions and with what the help row promises.

Issue #418, opened 2026-07-21
[Feature]: token-driven create form with live branch/path preview


kbrdn1:
## Problem

`CreateForm` hardcodes three fields (`Type`, `Issue`, `Desc`). Once `branch_pattern` is genuinely honoured, the form must follow the pattern rather than assume the default triple: a pattern without `{issue}` should not present an Issue field.

## Proposal

Drive the create form from the pattern's tokens: parse the pattern, produce an ordered token list, and generate one focusable field per token in pattern order.

- `{type}` -> cycler over the configured `branch_types`
- `{issue}` -> digits-only input
- `{desc}` -> text input
- free-form mode -> a single `Name` field

Plus, and this is the part that makes custom patterns actually usable: a **live preview line** showing the resolved branch and path as the user types, via `expand_placeholders`. It is cheap (the function already exists) and it is the only thing that makes a custom pattern legible without reading documentation.

`Ctrl-T` toggles Structured / Free-form.

## Scope note

A narrower version was considered: keep the three hardcoded field types and only show/hide them based on the tokens present. That covers the realistic patterns for roughly a third of the cost. The generic token-driven form was chosen instead, so exotic tokens are supported rather than silently dropped from the form.

Worth flagging: `CreateForm` is currently covered by state-machine tests (`tests/tui_state_create_form_tests.rs`), and this reshapes that state. Expect the test file to be substantially rewritten rather than extended.

## Tests

`tests/tui_state_create_form_tests.rs`: field generation per pattern shape, focus order matching pattern order, per-token validation, the free-form toggle, and preview-line correctness against `expand_placeholders`.

## Depends on

The pattern-to-regex derivation, which establishes the token model the form reads from.

kbrdn1:
Landed on `dev` in #492 (`b4c92bf`). Closing by hand: a pull request targeting `dev` does not fire `Closes`.

Two of the three asks here had already shipped ahead of this issue, so what actually landed is the field set and the focus order, derived from the patterns. The live branch/dir preview came with #217's follow-up and #416, and `Ctrl-T` came with #416.

[...]

Issue #417, opened 2026-07-21
[Feature]: derive the branch parser from branch_pattern (drop BRANCH_RE)


kbrdn1:
## Problem

The formatter is derived from `worktree.branch_pattern`; the parser is a separate hardcoded regex (`src/naming.rs:22`). They can disagree, and when they do, features that re-parse the branch name silently stop working. The interim guard turns that into a warning; this issue removes the cause.

## Proposal

Compile the pattern into a regex, and delete `BRANCH_RE`.

`{type}/#{issue}-{desc}` compiles mechanically:

- literal segments are escaped (`/`, `#`, `-`)
- each `{token}` becomes a named group with the charset it already implies:
  - `{type}` -> an alternation of the configured `branch_types` (`resolved_branch_types` already exists), which is strictly better than `[a-z]+` because a finite charset makes far more adjacencies unambiguous
  - `{issue}` -> `(?P<issue>\d+)`
  - `{desc}` -> `(?P<desc>[a-z0-9-]+)`

Result: round-trip for any pattern, one source of truth, and the level-1 capability the config has been advertising all along actually works.

## The separator constraint, and why the obvious rule is too weak

"Require a literal separator between adjacent tokens" is not sufficient. The correct rule is:

> **the separator must not belong to the charset of the token on its left**

Counter-example. `{desc}-{issue}` has a literal separator. But `desc = [a-z0-9-]+` **contains `-`**. On `user-auth-42`, the greedy group swallows `user-auth-4` and leaves `2` for `{issue}`.

The regex does not fail. It is deterministic. It just produces the **wrong split, silently**, which is the dangerous failure mode: not a parse error, a semantic surprise.

The current default is safe (`{type}/#{issue}-{desc}`: `/` and `#` are outside `[a-z]+`, `-` is outside `\d+`), so the constraint breaks nothing that exists today. It is validated mechanically at `gwm config validate` and deserves a test per adjacent-token combination.

## Alternative considered

A finite set of named, pre-tested patterns, each with a hand-written regex. Safer and trivially testable, but it **withdraws a freedom the config already advertises** (`branch_pattern` is a free-text field, editable from the Settings panel). Rejected for that reason.

## Contract

No `SCHEMA_VERSION` bump. `JsonWorktree` carries no `type` / `desc`; see the free-form issue for the full verification.

## Tests

[...]

Issue #416, opened 2026-07-21
[Feature]: free-form worktree naming (gwm create --name)


kbrdn1:
## Problem

`gwm create` requires the full `<type> <issue> <desc>` triple. `BranchSpec` has three mandatory fields, `desc` is validated against `DESC_RE`, and the CLI takes three positional arguments. There is no way to spin up a worktree that is simply named `spike-redis`.

Nobody has explicitly asked for a custom *pattern*, but "let me name it whatever I want" is the implicit ask behind it, and it is by far the cheaper half.

## Proposal

Introduce free-form naming as a distinct mode:

```rust
enum WorktreeName {
  Structured(BranchSpec),
  Freeform(String),
}
```

### CLI

A third entry point, exclusive with the existing ones via `conflicts_with`:

```
gwm create <type> <issue> <desc>   # unchanged, canonical, non-breaking
gwm create --name spike-redis      # free-form
```

Rejected: inferring the mode from the argument count ("1 arg means free-form, 3 args means structured"). Ambiguous and surprising.

Free-form names are sanitised for filesystem and git-ref safety only, not against `DESC_RE`.

### TUI

`Ctrl-T` toggles Structured / Free-form in the create form. The free-form mode shows a single `Name` field.

### Discovery, and this is the important part

A branch that does not match the structured pattern is **`Freeform`, not invalid**. `parse_branch` already returns `Option` and all six call sites already handle `None` (`src/github.rs:105`, `src/doctor.rs:591`, `src/lifecycle.rs:119`, `src/cli.rs:2450`, `src/cli.rs:3090`, `src/tui/app.rs:2988`). The work is making sure each treats `None` as "unstructured, leave it alone" rather than as an error.

### Degradation

Features keyed on `type` / `issue` become clean no-ops, never errors:

- issue/PR auto-linking does not fire; `gwm link` remains available for manual linking
- gitmoji falls back to a configurable default, or to nothing
- `doctor` treats a free-form branch as a user branch and does not flag it

## Contract

[...]

PR #472, opened 2026-07-27
🐛 fix(config): warn when branch_pattern is customised (parser ignores it)

### LINKED ISSUE #415: [Fix]: warn when branch_pattern is customised (parser ignores it)
## Problem

`worktree.branch_pattern` is configurable (`src/config.rs:291`, default `{type}/#{issue}-{desc}`) and it is editable **live from the TUI Settings panel** (`SettingField::WorktreeBranchPattern`, `src/tui/state/config_panel.rs:286,331,387`).

`BranchSpec::branch_name` honours it when **formatting** (`src/naming.rs:118-126`).

`parse_branch` does not. It matches a hardcoded regex (`src/naming.rs:22`):

```rust
static BRANCH_RE: LazyLock<Regex> =
  LazyLock::new(|| Regex::new(r"^([a-z]+)/#(\d+)-([a-z0-9-]+)$").expect("static BRANCH_RE compiles"));
```

So the formatter follows the configured pattern and the parser does not. A user who changes `branch_pattern` in the Settings panel silently breaks every feature that keys on re-parsing the branch name:

- issue/PR auto-linking (`src/github.rs:105`)
- gitmoji selection by type (`src/gitmoji.rs`)
- the branch-convention check in `doctor` (`src/doctor.rs:591`)
- lifecycle placeholder resolution (`src/lifecycle.rs:119`)

No error, no warning. Those features just quietly become no-ops, with no way for the user to connect the cause to the effect.

## Proposal

Interim guard, independent of any larger fix:

- `gwm doctor` and `gwm config validate` emit a warning when `worktree.branch_pattern` differs from the default
- the message names the consequence explicitly, for example: `branch_pattern differs from the default; structured parsing does not follow it, so issue/PR auto-linking, gitmoji and the branch-convention check will be inactive on branches created with this pattern`

This does not fix the bug. It converts a silent failure into a stated limitation, which is the honest state of affairs until the parser is derived from the pattern (separate issue).

## Note on `path_pattern`

`worktree.path_pattern` (`src/config.rs:290`) is configurable too, but it does **not** have the same problem: `worktree_dirname` is generation-only (`src/naming.rs:129`, called from `src/cli.rs:2217`, `src/tui/app.rs:3055,3277`), and `WorktreeInfo.name` comes from the directory basename rather than from re-parsing a pattern. Verified, no round-trip, no guard needed.

## Tests

`tests/doctor_tests.rs` and `tests/config_cli_tests.rs`: a repo with a custom `branch_pattern` produces the warning; one with the default does not.

kbrdn1:
## Description

`worktree.branch_pattern` is honoured when gwm **writes** a branch name (`BranchSpec::branch_name`) and ignored when it **reads one back** (`parse_branch` still matches the hardcoded `BRANCH_RE`). Customising the pattern — including live from the TUI Settings panel — therefore produces branches gwm can no longer decompose, and every feature keyed on the parsed segments silently becomes a no-op: issue/PR auto-linking, gitmoji selection, the doctor branch-convention check, lifecycle placeholders.

This PR does not fix that. It states it: `gwm doctor` and `gwm config validate` now name the pattern and its consequences, so the failure stops being silent.

Closes #415

## Type of change

- [x] 🐛 Fix (bug fix)
- [x] 📝 Docs

## Changes

- `naming::branch_pattern_warning(pattern) -> Option<String>` — the single predicate, returning the user-facing message when the pattern diverges from the default. #417 swaps its body for the real pattern-to-regex round-trip check without touching either call site.
- `gwm doctor` — new check `worktree.branch_pattern round-trips through the parser`, registered **last** in the report so the existing 1-8 numbering in the docs stays valid. Warning, not Failed: the config is valid and the worktrees it produces work, only the structured extras go quiet.
- `gwm config validate` — prints the same message on **stderr** and still exits `0`. `validate_rendered` / `validate_file` now return the parsed `Config` so `validate()` reads the resolved value without a second parse.
- `config::default_branch_pattern` lifted to `pub(crate)` so the comparison has a single source of truth.
- Docs: doctor page check #9 and a `.gwm.toml` schema section on the cost of customising, both EN + FR.

Explicitly out of scope, per the issue: no `path_pattern` guard (generation-only, `WorktreeInfo.name` comes from the directory basename — no round-trip to break), and no TUI change.

## Tests

- [x] `cargo test` passes locally (full suite, run under a CI-like minimal `PATH`)

Written red-first; both new doctor tests failed on `expected a 'branch_pattern' check in the report` and the config-validate test on an empty stderr before the implementation landed.

- `tests/doctor_tests.rs` — a custom pattern yields `CheckStatus::Warning` whose detail names all three consequences (`auto-linking`, `gitmoji`, `branch-convention`); the default yields `Ok`. Deliberately does **not** assert `report.exit_code()`, which is env-dependent (missing binaries also warn).
- `tests/config_cli_tests.rs` — a custom pattern exits `0` with the warning on stderr; the default exits `0` with no `branch_pattern` mention.

## Screenshots / TUI captures

n/a — no TUI change.

Verified against this repo with the built binary:

```
✓ worktree.branch_pattern round-trips through the parser
    default pattern — `parse_branch` reads back what `branch_name` writes
```

`gwm doctor` on gwm-cli itself stays exit `0`: `.gwm.toml` sets `branch_pattern` explicitly, but to the default value, so the CI advisory doctor job is unaffected.

## Checklist

- [ ] README updated if CLI surface changed — no new command or flag
- [ ] `examples/gwm.toml.example` updated if config schema changed — no schema change, the key already existed

## Linked issues / docs

- Issue: #415
- Follow-ups in the same sequence: #416 (free-form `--name`), #417 (derive the parser from the pattern, removes the cause), #418 (token-driven create form)

## Notes for reviewers

Two things worth a look:

1. **Check ordering.** The new check is registered after `check_tui_keymap` rather than next to the other config checks, purely so `docs/5.integrations/2.doctor.md` keeps its 1-8 numbering. Say the word if you'd rather have it grouped logically and pay the doc renumber.
2. **`validate_rendered` now returns `Config`.** `write_and_validate` matches on `Ok(_)` instead of `Ok(())`; nothing else consumed the value. The alternative was a second `toml::from_str` in `validate()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,15 +13,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `gwm doctor` and `gwm config validate` now warn when
-  `worktree.branch_pattern` differs from the default `{type}/#{issue}-{desc}`.
+  `worktree.branch_pattern` does not survive a format-then-parse round-trip.
   The pattern drives how a branch name is *written* but not how one is *read
-  back* — the parser is still a hardcoded regex — so customising it silently
-  disabled issue/PR auto-linking, gitmoji selection and the branch-convention
-  check. The warning names those consequences instead of leaving the user to
-  discover them. `gwm config validate` prints it on stderr and still exits `0`:
-  a custom pattern is valid configuration, not an error. This states the
-  limitation, it does not remove it — deriving the parser from the pattern is
-  tracked by #417. ([#415](https://github.com/kbrdn1/gwm-cli/issues/415))
+  back* — the parser is still a hardcoded regex — so a mismatched pattern
+  silently broke issue/PR auto-linking, gitmoji selection, lifecycle hook
+  placeholders and the branch-convention check. The warning names whichever
+  segments actually break: a custom pattern is not automatically a broken one
+  (`{type}/#{issue}-prefix-{desc}` still recovers `type` and `issue`, only
+  `desc` comes back wrong), and claiming otherwise would defeat the point of a
+  warning whose whole value is accuracy. `gwm config validate` prints it on
+  stderr, reads the *effective* pattern so one set only in the global
+  `~/.config/gwm/config.toml` is caught too, and still exits `0`: a custom
+  pattern is valid configuration, not an error. This states the limitation, it
+  does not remove it — deriving the parser from the pattern is tracked by #417.
+  ([#415](https://github.com/kbrdn1/gwm-cli/issues/415))
 
 ### Docs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   segments actually break: a custom pattern is not automatically a broken one
   (`{type}/#{issue}-prefix-{desc}` still recovers `type` and `issue`, only
   `desc` comes back wrong), and claiming otherwise would defeat the point of a
-  warning whose whole value is accuracy. `gwm config validate` prints it on
+  warning whose whole value is accuracy. The probe enumerates the value space
+  `gwm create` admits rather than sampling it — every configured branch type,
+  a single- and a multi-digit issue, a desc with and without the `-` it
+  allows, and the real repo name for `{repo}` — so a pattern that breaks on
+  only some values is reported as breaking on only some values. `gwm config validate` prints it on
   stderr, reads the *effective* pattern so one set only in the global
   `~/.config/gwm/config.toml` is caught too, and still exits `0`: a custom
   pattern is valid configuration, not an error. This states the limitation, it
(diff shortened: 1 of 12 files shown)


kbrdn1:
Third note for reviewers, found while verifying the two consumers agree — they don't, in one reachable case.

`Config::merge_layered` layers the **whole** `Config`, so `[worktree].branch_pattern` can be set in the global `~/.config/gwm/config.toml`. `gwm doctor` reads the merged `ctx.config` and warns. `gwm config validate` reads the repo `.gwm.toml` only and stays quiet.

So a user who customises the pattern globally and leaves the repo file alone gets the warning from `doctor` and silence from `config validate`.

Not changed here, deliberately: `gwm config validate` validates one file by design — its success line names that path (`<path> is valid`) — and making it merge the global layer is a change to what the command *means*, well outside "state the limitation". Worth its own issue if the divergence bothers you; `doctor` is the merged-view command and it does the right thing.

kbrdn1:
Follow-up filed: #473 — `gwm config list` / `get` echo config values raw on `dev` too. Pre-existing, deliberately left out of this PR rather than widening its scope; the new echo path this PR adds is sanitised here.

Issue #334, opened 2026-06-19
[Bug]: help overlay (?) omits the exec / clean overlay bindings (x / X)


kbrdn1:
## Description

The TUI help overlay (`?`) is hand-curated in `ui::help_rows`. The #325 exec (`x`) / clean (`X`) overlays were added to the keymap (`gwm tui keys` lists them), the palette, and the modal contexts, but **not** to the curated `help_rows` list — so the in-app `?` overlay doesn't document them. No test enforces help-overlay completeness, so CI stayed green.

## Fix
- Add `Action::ExecOverlay` / `Action::CleanOverlay` entries to `help_rows` (List View section, next to `command_logs` / `config_panel`).
- Add a completeness test so a future Action can't silently miss the help overlay.

Follow-up to #325 / PR #333.

kbrdn1:
Completeness guard landed in PR #347 (merged into `dev`). `help_overlay_documents_every_action` iterates `Action::all()` and asserts each action's default chord is documented in the `?` overlay, so a future Action can't silently miss it. The exec/clean help entries themselves shipped in v1.0.0.

PR #474, opened 2026-07-28
✨ feat(cli,tui): free-form worktree naming (gwm create --name)

### LINKED ISSUE #416: [Feature]: free-form worktree naming (gwm create --name)

kbrdn1:
## Description

`gwm create` required the full `<type> <issue> <desc>` triple, and not every worktree corresponds to an issue. `gwm create --name spike-redis` skips the convention; in the TUI, `Ctrl-T` toggles the create form between the structured triple and a single `Name` field.

Closes #416

## Type of change

- [x] 📝 Docs

## Changes

- `WorktreeName::{Structured, Freeform}` in `naming.rs`, with `branch_name` / `worktree_dirname` / `worktree_path` dispatching on the variant.
- `gwm create --name`, exclusive with the positionals via clap (`conflicts_with_all` one way, `required_unless_present` the other) so a partial triple stays the typo it always was.
- TUI: `Ctrl-T` → `toggle_mode`, single `Name` field, mode named in the overlay title, type selector and issue field dropped in free-form rather than shown inert.
- Two refusal messages reworded — a free-form branch now reaches them by design, not by mistake.
- Docs EN + FR: CLI reference section, TUI keybinding row, CHANGELOG.

### Decisions the issue left open

**Patterns do not apply to a free-form name.** `branch_pattern` / `path_pattern` are written in terms of `{type}` / `{issue}` / `{desc}` and a free-form name has none of them, so the name *is* the branch and the directory is the name with `/` flattened to `-` — the same relationship the default pattern pair already has (`feat/#42-x` lives at `feat-42-x`). `[worktree].base` still applies, so free-form worktrees land beside the structured ones.

**`gwm commit-prefix` stays an error** rather than falling back. The issue suggested "a configurable default, or nothing"; that command exists solely to derive a prefix from the branch *type*, so there is no honest default, and a new config key for it is something nobody has asked for. The message now names the missing shape instead of implying the branch is malformed.

**Validation is delegated, not hand-rolled.** Ref legality goes through libgit2's own `Reference::is_valid_name` against the real `refs/heads/<name>`. Two rules sit outside it, kept separate so the reasons stay honest:
- `.` / `..` as a path component — git accepts `..` inside a longer component, but a worktree directory named `..` would escape the base.
- a leading `-` — libgit2 accepts `refs/heads/-x`, **verified** (the test caught my wrong assumption), but the name would be unusable as an argument to `gwm remove`, `git branch -d` or `cd`.

### Degradation

Confirmed at each call site rather than assumed. The `parse_branch` sites needed no `None` handling added — they already treat it as "unstructured, leave alone", including the doctor orphan check.

| Feature | On a free-form worktree |
|:--------|:------------------------|
| issue auto-linking | inactive — `gwm link --issue <N>` remains |
| **PR/MR detection** | **still works** — `Forge::find_pr_for_branch` queries with the whole branch name and never parses it |
| gitmoji / `gwm commit-prefix` | errors, with a message naming the missing shape |
| `doctor` orphan check | user-managed branch, never flagged |
| hook placeholders (remove / bootstrap) | resolve empty — the honest value |
| TUI edit form | says the form rebuilds the triple; rename with git |

## Tests

- [x] `cargo test` passes locally (full suite, CI-like env: `GWM_NO_GLOBAL_CONFIG=1` + minimal `PATH`)

Written red-first at each layer:

- `naming_tests.rs` — the enum, pattern non-application, `/` flattening, and a rejection table driven through the git oracle. The `-leading` case went green when it should have been red, which is how the wrong assumption surfaced.
- `cli_binary.rs` — `--name` in help, partial-triple rejection, flag conflict, the happy path asserting `path_pattern` was *not* applied, and a rejected name.
- `tui_state_create_form_tests.rs` — toggle, focus landing, buffer preservation across a round trip, rotation no-op, length cap, reset.
- `tui_modal_render_tests.rs` — free-form overlay shows `Name` and does **not** render `Type` / `Issue`.

The help-overlay completeness guard (#334) went red on the new modal verb, as designed.

## Screenshots / TUI captures

None attached — the render is covered by a buffer assertion instead. **This PR touches the TUI, so it needs your local install-and-validate pass before merge.**

## Checklist

- [x] README updated if CLI surface changed — the CLI reference under `docs/` carries the flag; README does not enumerate flags
- [ ] `examples/gwm.toml.example` updated if config schema changed — no schema change

## Linked issues / docs

- Issue: #416
- Next in the sequence: #417 (derive the parser from the pattern), which #472 left an acceptance table on; then #418 (token-driven create form), which reshapes the form this PR just extended

## Notes for reviewers

Worth a look:

1. **`MAX_NAME_LEN` is now `naming::MAX_DIR_COMPONENT_BYTES` (255), counted in bytes.** It started at 200 characters, mirroring `MAX_DESC_LEN`, which meant the form silently truncated a 201-character name the validator would have accepted — submitting a different branch than the one typed. Review pass 3 caught it; the form's cap and the validator's are the same constant now, so they cannot drift apart again.
2. **`#418` will rewrite `create_form.rs`.** That issue's own body warns the file will be substantially rewritten. This PR adds a mode dimension the token-driven form will have to carry; the `Mode` enum should survive, the `Field` enum probably will not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,10 +19,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   positionals, so a partial triple is still the typo it always was; the mode
   is chosen explicitly, never inferred from how many arguments arrived.
 
-  The name becomes the branch verbatim. `branch_pattern` / `path_pattern` do
-  not apply — they are written in terms of `{type}` / `{issue}` / `{desc}`,
+  The name becomes the branch verbatim, and is validated verbatim: `--name
+  " spike"` is refused rather than trimmed into `spike`, which would be a
+  different branch from the one asked for. `branch_pattern` / `path_pattern`
+  do not apply — they are written in terms of `{type}` / `{issue}` / `{desc}`,
   and a free-form name has none of them — while `[worktree].base` still does,
-  so free-form worktrees land beside the structured ones.
+  so free-form worktrees land beside the structured ones. `base` is only
+  expanded with the placeholders it documents, though: the structured path
+  feeds `{type}` / `{issue}` / `{desc}` through `base` too, and since an
+  unfed placeholder is left *literal*, a base written with one of them is
+  refused here rather than turned into a directory called `{type}`.
 
   What a free-form worktree gives up is stated rather than discovered: issue
   auto-linking goes inactive (`gwm link` remains), `gwm commit-prefix` errors
@@ -32,9 +38,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   treats the branch as user-managed and never flags it.
 
   Names are validated against libgit2's own branch-name rules rather than a
-  hand-written list, plus two rules of our own: no `.` / `..` path component
-  (a worktree directory named `..` would escape the base) and no leading `-`
-  (git accepts it; `gwm remove` and `git branch -d` would not).
+  hand-written list, plus three rules of our own: no `.` / `..` path component
+  (a worktree directory named `..` would escape the base), no leading `-`
+  (git accepts it; `gwm remove` and `git branch -d` would not), and a 255-byte
+  cap — a branch name is a *path* of components while the worktree directory
+  is a single one, so `a×130/b×130` is a legal ref and an illegal directory
+  name, and without the cap the branch is created before the directory fails,
+  leaving it orphaned.
   No `SCHEMA_VERSION` bump — `JsonWorktree` carries no `type` / `desc`, so
   the wire format is unchanged.
   ([#416](https://github.com/kbrdn1/gwm-cli/issues/416))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Free-form worktree naming. `gwm create --name spike-redis` skips the
+  `<type> <issue> <desc>` triple entirely — not every worktree corresponds to
+  an issue. In the TUI, `Ctrl-T` toggles the create form between the
+  structured triple and a single `Name` field. The flag is exclusive with the
+  positionals, so a partial triple is still the typo it always was; the mode
+  is chosen explicitly, never inferred from how many arguments arrived.
+
+  The name becomes the branch verbatim. `branch_pattern` / `path_pattern` do
+  not apply — they are written in terms of `{type}` / `{issue}` / `{desc}`,
+  and a free-form name has none of them — while `[worktree].base` still does,
+  so free-form worktrees land beside the structured ones.
+
+  What a free-form worktree gives up is stated rather than discovered: issue
+  auto-linking goes inactive (`gwm link` remains), `gwm commit-prefix` errors
+  because a prefix is derived from the branch type and there is none, and
+  remove/bootstrap hook placeholders resolve empty. PR/MR detection is
+  unaffected — it queries the forge with the whole branch name. `doctor`
+  treats the branch as user-managed and never flags it.
+
+  Names are validated against libgit2's own branch-name rules rather than a
+  hand-written list, plus two rules of our own: no `.` / `..` path component
+  (a worktree directory named `..` would escape the base) and no leading `-`
+  (git accepts it; `gwm remove` and `git branch -d` would not).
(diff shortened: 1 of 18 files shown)


kbrdn1:
## Review pass 2 — one finding dismissed, on the record

`P0P1P2_COUNT` went 4 → 4 across two passes, and three of the second batch were **one class**: a free-form name widens the input space of consumers that `DESC_RE` (`[a-z0-9][a-z0-9-]*`) used to keep narrow. Rather than keep patching examples, the rules are now enumerated from the three things a free-form name has to be at once — a git branch, a single filesystem path component, a literal value during hook expansion — plus the one CLI-ergonomics rule (`-` in front). That is in `WorktreeName::freeform`'s doc comment, so the next reader inherits the invariant rather than the list.

**Dismissed and deferred: Windows path validation** → filed as #475.

`< > " |` and reserved device names (`CON`, `NUL`, `COM1`…) pass every check here and still cannot become a directory on Windows, so such a name fails inside `worktree::add` rather than up front — after `pre_create` hooks, leaving the branch behind. Real, same late-failure shape as the 255-byte cap fixed earlier in this PR, and only reachable through free-form naming.

Not fixed here for two reasons, both about verification rather than effort:

1. **It cannot be measured from this machine.** What libgit2 0.21 already refuses on Windows is not knowable from a Mac, so the residual rule set would be guessed. Every other rule in this PR was reproduced against a real binary first — the orphaned branch, the literal `{type}` directory, the silent trim. This one could not be.
2. **A `#[cfg(windows)]` rule is exercised by one runner only**, which is exactly what [CLAUDE.md's rule on pre-validating environment-dependent tests](../blob/dev/CLAUDE.md) argues against.

#475 carries the full analysis, including the observation that `worktree::add` creating the branch *before* the directory is what turns any late failure into an orphaned branch — blunting the class rather than this one input set.

If a later pass raises it again, this is the answer.

Issue #304, opened 2026-06-15
[Feature]: workspace mode — orphan linked worktrees + activation-failure action guard


kbrdn1:
## Problem

Follow-up to #36 (multi-repo workspace mode, PR #303). Two nuanced
correctness edge cases surfaced by the Codex review loop, deferred from the
PR because each is a design decision rather than a contained fix:

### 1. Standalone linked worktrees under the root are dropped

\`workspace::discover\` currently skips any \`repo.is_worktree()\` entry to avoid
duplicate rows when a repo *and* one of its linked worktrees are both direct
children of the root (the dedup the PR shipped). But if the root is a
directory of linked worktrees whose **main checkouts live elsewhere**, those
are usable git checkouts and the unconditional skip omits them (or reports an
empty workspace).

The dedup rationale only holds when the owning main repo is *also* discovered
under the same root. Desired behaviour: skip a linked worktree only when its
owner is present under the root; otherwise include it (or resolve it to its
owner). Needs a design decision on include-vs-resolve.

### 2. No action guard after a failed repo activation

\`App::sync_active_repo\` re-opens the selected row's repo before each frame.
If that \`Repository::open\` fails (repo moved/deleted/permissions/corrupt
\`.git\` after listing), the error path only sets the status line and leaves
\`self.repo\`/\`workdir\`/\`config\` pointing at the **previously** active repo while
the selected row belongs to the failed one. A subsequent mutation that
combines selected-row data with \`self.repo\` (delete, rename, bootstrap, link)
could then target the wrong repository.

(Partly mitigated today: the next workspace auto-refresh re-lists all repos
and drops the dead one, bounding the mismatch window — but a mutation in the
interim is still a wrong-target risk.)

Desired behaviour: disable mutating actions for the selected row (or snap the
selection back to a still-valid repo) until activation succeeds.

## Surface

- \`src/workspace.rs\` — \`discover\` linked-worktree handling.
- \`src/tui/app.rs\` — \`sync_active_repo\` failure path + mutating-action guard.

## Backward compatibility

No breaking change; both are hardening of workspace mode (#36).

## Additional context

Surfaced by the \`/me:loop:codex-review-pr\` loop on PR #303 (rounds 4–5).
Blocking findings (P0/P1) were zero on every round; these are the residual
P2s deferred as design-level follow-ups.

kbrdn1:
[...]

PR #476, opened 2026-07-28
✨ feat(naming): derive the branch parser from worktree.branch_pattern

### LINKED ISSUE #417: [Feature]: derive the branch parser from branch_pattern (drop BRANCH_RE)

kbrdn1:
## Description

`worktree.branch_pattern` decided how gwm **writes** a branch name; a hardcoded `^([a-z]+)/#(\d+)-([a-z0-9-]+)$` decided how it **reads** one. A repo that set `branch_pattern = "{type}-{issue}-{desc}"` created `feat-41-foo` and then failed to recognise the branch it had just created.

`BranchParser` compiles the pattern into the regex that reads it back. `BRANCH_RE` is gone, and with it `parse_branch`, which had no production caller left.

Closes #417

## Type of change

- [x] 🐛 Fix (bug fix)

## Changes

- `naming::BranchParser` compiles `branch_pattern` into a parser, mirroring the exact `expand_placeholders` call `BranchSpec::branch_name` makes. `{type}` / `{issue}` / `{desc}` become named groups, `{repo}` / `{home}` become escaped literals, everything else stays literal (which is correct for `{repo_path}` / `{repo_parent}`, since the branch path passes `None` for `repo_path`).
- Threaded through every consumer: `github::read_link` (auto-linking), `worktree::list` (one compile per listing, not per branch), `doctor` (orphan check), `lifecycle::HookContext::for_worktree` (remove / bootstrap placeholders), `cmd_pr` (`[pr_template.by_type]` + body placeholders), the TUI rename, and `gwm commit-prefix`.
- `branch_pattern_warning` keeps both call sites and reports the residue: a pattern the compiler refuses, and a segment the pattern never writes.
- Docs table (EN + FR) rewritten from the real check.

## Two corrections to the issue as filed

**1. `{desc}-{issue}` is not ambiguous.** The issue proposed rejecting any pattern whose separator belongs to the left placeholder's charset, citing `user-auth-42` splitting as `user-auth-4` + `2`. Measured against the engine that ships, it does not: greedy repetition backtracks, the literal `-` has to match after the group, and since `{issue}` is `\d+` it cannot contain the separator, so exactly one split is valid. Implementing the rule as written would have refused a pattern that works. Full measurement in [the correction comment](https://github.com/kbrdn1/gwm-cli/issues/417#issuecomment-5103050955).

The rule that ships is narrower: **adjacency**. Two placeholders with no literal between them are refused (`{issue}{desc}` reads `42123-x` back as `4212` + `3-x`; `{desc}{issue}` is ambiguous outright, since `a12` is what both `a` + `12` and `a1` + `2` produce), plus the same placeholder twice.

**2. The probe is not replaced by the compiler.** Adjacency-rejection is necessary, not sufficient, so `branch_pattern_warning` still probes the *derived* parser as a backstop. That earns its keep, measured rather than assumed: a `~`-leading pattern does not round-trip, because the formatter ends with `shellexpand::tilde` and the reader has no way to undo it.

## No regression: the 1.5.0 parity table is the acceptance criterion

The set of branches 1.5.0 could read is exactly the set matching its `^([a-z]+)/#(\d+)-([a-z0-9-]+)$`, read verbatim off the `v1.5.0` tag. Computed rather than reasoned about, that is six patterns:

| Pattern | 1.5.0 read | now |
|:--|:--|:--|
| `{type}/#{issue}-{desc}` | `feat / 42 / my-desc` | same |
| `feat/#{issue}-{desc}` | `feat / 42 / my-desc` | same (type recovered from the literal) |
| `{type}/#1-{desc}` | `feat / 1 / my-desc` | same (issue recovered) |
| `{type}/#{issue}-fixed` | `feat / 42 / fixed` | same (desc recovered) |
| `{type}/#{issue}-prefix-{desc}` | `feat / 42 / prefix-my-desc` | `feat / 42 / my-desc` (1.5.0 was wrong, #415 said so) |
| `{type}/#{issue}-{desc}-{repo}` | `feat / 42 / my-desc-gwm-cli` | `feat / 42 / my-desc` (same) |

`every_branch_1_5_0_could_read_is_still_read_the_same_way` pins the first four; `the_two_patterns_1_5_0_read_wrongly_are_read_correctly_now` pins that matching 1.5.0 on the last two would mean keeping the bug.

## Frozen segments are recovered, not dropped

`feat/#{issue}-{desc}` hardcodes the type; 1.5.0 read it back because its regex happened to have a group where the literal sits. The parser recovers such literals as constants, so gitmoji, auto-linking and `gwm commit-prefix` keep working.

The recovery is an exact match, not an inference:

- **type** : looked up in the repo's configured branch types. `feature/{issue}-{desc}` recovers nothing, because `feature` names a namespace and is not in the list.
- **issue** : all digits.
- **desc** : whatever `DESC_RE` accepts, run last on what the stricter two left. That is what keeps `feat/#1-fixed` from reading its description as `1-fixed`.

A segment is claimed only when exactly one candidate survives. `{repo}` and `{home}` are deliberately not sources: the user wrote `feat` in the pattern and meant it, but nobody chose the repo's name for this, and a repo called `docs` must not type its own branches.

`desc` is the loose oracle and it self-reports rather than hiding. In `wt/{type}/#{issue}` the namespace `wt` is the one `DESC_RE`-shaped token, so it is claimed, and the round-trip probe immediately says the description does not match what `gwm create` was given. Every constant that costs something surfaces that way, so no separate disclosure channel was needed.

What a frozen segment costs is reported exactly as #415 reported it: `gwm create fix 42 x` under `feat/#{issue}-{desc}` writes a `feat/` branch, so the type asked for is not the one read back, and `gwm doctor` warns. With `feat` as the only configured type there is no loss and it stays quiet, which is the case #415's own review pinned.

## `{type}` stays `[a-z]+`

The issue proposed compiling `{type}` into an alternation of the configured branch types. That would stop recognising a branch created before a type was retired from `.gwm.toml`, taking `doctor`'s orphan check and `gwm commit-prefix` away from a name the previous release read fine, and collapsing the TUI rename's precise "type is not configured" message into a generic one.

Nothing needs it: once adjacent placeholders are refused, `[a-z]+` splits every pattern in the documented table, and the one consumer that genuinely requires a *configured* type, the TUI rename, checks the resolved list itself.

## `gwm commit-prefix` guard

A pattern that carries no `{type}` or `{issue}` *and* freezes neither (`{type}/{desc}`, `{issue}-{desc}`) parses perfectly and yields an empty segment, so `resolve_prefix` would have printed ` (#):` straight into a commit message. That is now an error naming the placeholder to add, distinct from the #416 free-form message. Verified end to end against three real repos: `feat/#{issue}-{desc}` renders `:sparkles: feat(#42):`, `{type}/#1-{desc}` renders `:bug: fix(#1):`, and `{type}/{desc}` errors.

## Tests

- [x] `cargo test` passes locally (under `GWM_NO_GLOBAL_CONFIG=1` and a CI-like minimal `PATH`)

Driven from a CLI-level red test (`commit_prefix_reads_a_branch_written_with_a_custom_pattern`) rather than from the compiler outward, because the acceptance criterion is "the features keep working", not "the regex is right".

Added: per-token compilation, the plausible-convention table round-tripping, both ambiguity refusals, the type narrowing, and a guard that reads the token list **out of `expand_placeholders` itself** so a placeholder added later cannot silently fall through as a literal and reintroduce the divergence this PR deletes.

Rewritten: the #415 block. Ten tests asserted that specific patterns were broken; they are not any more. Each keeps its original reasoning as history and asserts the new truth. Two needed a surviving vehicle rather than an inversion:

- `control_characters_in_the_pattern_never_reach_the_terminal` is a security test (P1 from #415's review) and must not weaken to "no longer warns". #417 added a third echoing path and moved the other two, so all three are exercised now: the compile error, the missing-placeholder verdict, and the probe's `e.g.` example.
- `the_consumer_mapping_matches_the_call_sites` pins "PR/MR detection is unaffected" and "remove/bootstrap, not lifecycle". Re-anchored on the pattern that still reaches that branch.

## Notes for reviewers

- The `desc` oracle is the loose one. `wt/{type}/#{issue}` reads `wt` as the description of every branch it writes. It is reported by the probe rather than silent, and the fix is to write `{desc}`, but it is the one place recovery guesses.
- `worktree::list` now reads `.gwm.toml` once per call (best-effort, `unwrap_or_default`) to compile the parser. `read_link`'s public signature is unchanged and derives per call; `read_link_with` is the hoisted form for loops.
- A repo that changes `branch_pattern` will have existing branches created under the old one. Those stop parsing, which is correct but worth stating: `doctor` reclassifies them as user-managed, and `commit-prefix` errors on them.

## Linked issues / docs

- Issue: #417
- Follows: #415 (PR #472), #416 (PR #474)
- Next: #418


kbrdn1:
### Review pass 5: one finding fixed, one deferred to #478

**Fixed (`41aa689`).** Third finding in three passes on the same claim: that `compile` tokenises the way `expand_placeholders` substitutes. Each pass found a shape the previous fix did not cover: a brace before a placeholder, an expansion carrying another token, and now a token formed across an expansion boundary (`{{repo}}` in a repo called `type` writes `feat/#42-x`, because the formatter substitutes what its own expansions produce).

Patching the third instance would have invited a fourth, so the claim is verified rather than argued: `compile` writes one probe branch through the real formatter and refuses the pattern when the parser it just built cannot read it back. The special-case scan added in pass 4 is gone, subsumed by this. A `~` prefix stays excluded on purpose, since that divergence is known and `branch_pattern_warning`'s probe already names every feature it takes down.

**Deferred to [#478](https://github.com/kbrdn1/gwm-cli/issues/478).** The other rename finding, a segment frozen in `branch_pattern` but still carried by `path_pattern`, where the rename recomputes the path from the branch and loses the value the directory held.

It predates this PR, verified against the tag rather than assumed:

- `git show v1.5.0:src/tui/app.rs` : `submit_edit_worktree` already computed `new_path` from `spec.worktree_path(...)` with no preservation, the same three lines.
- `git show v1.5.0:src/naming.rs` : `BRANCH_RE` is `^([a-z]+)/#(\d+)-([a-z0-9-]+)$`, so `feat/#{issue}-{desc}` + the default `path_pattern` already opened the form and moved `fix-42-x` to `feat-42-…` on any rename.

What #417 changes is reach, not behaviour. Fixing it properly means deriving a second parser from `path_pattern` to read the old directory back, then deciding what a rename means when the branch and the path disagree, which changes the shape of the rename and overlaps #418's live branch/path preview.

This PR does close the sharper half: a frozen segment the user *edits* is refused outright (pass 4), because changing it renamed nothing and moved the directory instead.

kbrdn1:
### Review pass 8: one fixed, three dismissed with evidence

This is the review loop's iteration cap, so it is the last fixing round on this PR.

**Fixed (`6a8c9cc`), the rename form used two different repo names.** In a workspace, two repos whose directories share a basename are disambiguated for display (the second becomes `api-2`, #304) and `App::repo_name` holds that name. Every formatter call in the rename flow already expands `{repo}` with it, but the parser #417 added was compiled from the workdir basename. Under a `{repo}` pattern they disagreed, so the form refused to open on a worktree the repo owns, reporting a branch it had written itself as free-form. One name now, which is the whole point of the issue.

**Dismissed: removing `parse_branch` is not a breaking change.** `src/lib.rs` opens with *"Internal test seam, NOT a public API"*, carries `#![doc(hidden)]`, and states that every `pub` item below has **no SemVer guarantee**. The stable contracts are named there and frozen by `tests/contract_tests.rs`: the CLI, the `--format=json` / daemon payloads, and the `.gwm.toml` schema. The lib ships only as a byproduct of shipping the binary.

**Dismissed: the frozen-segment / `path_pattern` interaction is [#478](https://github.com/kbrdn1/gwm-cli/issues/478).** Reported again this pass; the deferral was evidenced against the v1.5.0 tag in the pass-5 comment above. Fixing it means deriving a second parser from `path_pattern` and deciding what a rename means when the two patterns disagree.

**Dismissed here: the `git branch -d` suggestion in `gwm doctor`.** `format!("git branch -d {}", b)` is byte-for-byte identical in `git show v1.5.0:src/doctor.rs:664`; #417 does not touch that line. Reaching it with a name carrying shell metacharacters needs those characters to come from the pattern itself, since no capture charset admits them, i.e. it needs an untrusted `.gwm.toml`, which is a different threat model with its own track. Not folded into this PR.

PR #486, opened 2026-07-31
🐛 fix(naming): validate a free-form worktree name against Windows path rules

### LINKED ISSUE #475: Free-form worktree names are not validated against Windows path rules
## Context

#416 added `gwm create --name <name>` (PR #474). `WorktreeName::freeform` validates the name against three consumers: git's own branch-name oracle, a single filesystem path component (≤ 255 bytes, no `.` / `..`), and placeholder expansion (no `{` / `}`).

One class is deliberately left out of that PR: **Windows-specific path rules.**

## Problem

A name can pass every current check and still be impossible to create as a directory on Windows:

- characters git allows in a ref but Windows forbids in a path: `<`, `>`, `"`, `|`
- reserved device names, case-insensitively and including with an extension: `CON`, `PRN`, `AUX`, `NUL`, `COM1`–`COM9`, `LPT1`–`LPT9`, and forms like `CON.txt`
- a trailing space (git already refuses a trailing dot)

`gwm create --name 'foo|bar'` therefore fails inside `worktree::add` — after `pre_create` hooks have already run, and after the branch has been created, which leaves it orphaned. That is the same late-failure shape the 255-byte cap fixed in PR #474, just on a platform-specific input set.

This is only reachable through free-form naming: `DESC_RE` (`[a-z0-9][a-z0-9-]*`) keeps structured names out of it.

## Why it was deferred rather than fixed in #474

Two reasons, both about verification rather than effort:

1. It cannot be verified from a Unix machine. What libgit2 0.21 already rejects on Windows is not knowable from here, so the residual set would be guessed rather than measured.
2. A `#[cfg(windows)]` rule is a rule only one CI runner exercises, which the repo's own house rule on pre-validating environment-dependent tests argues against.

## Suggested shape

Either validate unconditionally on all platforms (a branch travels to teammates' machines through the forge, so a name that cannot be checked out on Windows is a cross-platform hazard, and the rule then runs on every runner), or gate it to `#[cfg(windows)]` with the test suite exercising the predicate directly rather than through the filesystem. The first is testable everywhere; the second is more permissive locally. Measure what libgit2 already refuses on Windows before writing the residual list.

Independently of which is chosen, `worktree::add` creating the branch before the directory is what turns a late failure into an orphaned branch — rolling the branch back when `repo.worktree` fails would blunt the whole class, not just this input set.

## Acceptance

- [ ] a name that Windows cannot host as a directory is refused before anything is created
- [ ] the residual rule set is derived from what libgit2 does *not* already reject, measured rather than assumed
- [ ] the predicate is unit-tested without needing a Windows filesystem
- [ ] `docs/3.cli/1.reference.md` + `docs/fr/3.cli/1.reference.md` drop the caveat added by #474

kbrdn1:
## Description

`WorktreeName::freeform` validated a name against three consumers: git's branch oracle, a single filesystem path component, and placeholder expansion. One of those consumers has a half the developing machine is not. `gwm create --name 'foo|bar'` or `--name CON` passed every check and failed inside `worktree::add`, after `pre_create` hooks had run and with the branch already created, leaving it orphaned. Same late-failure shape the 255-byte cap closed in #474, on an input set only free-form naming can reach (`DESC_RE` keeps structured names out of it).

Closes #475

## Type of change

- [x] 🐛 Fix (bug fix)
- [x] 📝 Docs
- [x] ✅ Tests
- [ ] ⚡ Perf
- [ ] 🔧 Chore / CI / build

## Changes

- `WINDOWS_FORBIDDEN_CHARS` and `WINDOWS_RESERVED_STEMS` in `src/naming.rs`, checked at the end of `WorktreeName::freeform` so a name failing git's rules too gets git's message.
- The reserved check walks `/`-separated segments rather than the flattened directory name alone.
- EN and FR CLI references: the "not validated yet" warning block becomes part of the accepted-name rules.
- `CHANGELOG.md`: the #416 entry no longer says Windows rules are deliberately uncovered, and #475 gets its own `Fixed` entry.

## The residual list is measured, not copied

The issue asked for the rule set to be derived from what libgit2 does *not* already reject. Every candidate was run through `git2::Branch::name_is_valid` on this machine before a line was written, and every new test carries that as a precondition assertion, so a reviewer can see that only gwm's own check can stop the case:

| input | `Branch::name_is_valid` | rule needed |
|---|---|---|
| `spike<x` `spike>x` `spike"x` `spike\|x` | accept | **yes** |
| `spike:x` `spike*x` `spike?x` `spike\x` | REJECT | no |
| `spike ` ` spike` `spi ke` `spike.` | REJECT | no |
| `spike\u{1}x` `spike\u{7f}x` `spike\tx` | REJECT | no |
| `CON` `con` `CON.txt` `NUL` `COM1` `LPT1` `COM¹` | accept | **yes** |
| `COM0` `CONX` `CON-x` `console` | accept | no, not reserved |

Win32 forbids nine characters in a path component (`< > : " / \ | ? *`). Four are already refused by git, and `/` is the ref separator `worktree_dirname` flattens to `-`, which leaves exactly four. Windows also refuses a trailing space or period, both of which git refuses first, so neither gets a rule of its own; `the_trailing_space_and_period_rules_are_left_to_git` pins that as a decision rather than an omission.

The reserved device names are what git knows nothing about. The list is verbatim from [Naming Files, Paths, and Namespaces](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file): `COM0` and `LPT0` are deliberately absent because they are absent there, and the ISO 8859-1 superscripts `¹ ² ³` are present because Windows reads them as digits in a device name. The comparison is on the stem before the first `.`, since that page documents `NUL.tar.gz` as equivalent to `NUL`, and case-insensitive because Windows path comparison is.

## Two decisions worth a reviewer's attention

**Unconditional, not `#[cfg(windows)]`.** A branch reaches a teammate through the forge, so a name no Windows checkout can host is a cross-platform hazard rather than a local one. Gating it would also leave a rule that a single CI runner ever exercises, which is the shape CLAUDE.md's guidance on pre-validating environment-dependent tests argues against. The timing argument is the one that makes this a pre-cut fix rather than a follow-up: `--name` merged into `dev` on 2026-07-28 and v1.5.0 was tagged 2026-07-26, so the flag has never shipped to a stable line and tightening it now is additive. After it ships, the same tightening is a breaking change.

**Per ref segment, not only the flattened dirname.** A loose ref is a file at `.git/refs/heads/<name>`, so every `/`-separated segment is a path component there too. `spike/CON` flattens to the perfectly legal directory `spike-CON` and is still a ref Windows cannot write. This half is derived from the Win32 spec and from how git stores loose refs, not measured on Windows, since no Windows machine was available; the character and device lists themselves are from the primary source above.

## Tests

- [x] `cargo test` passes locally (2486 passed, 0 failed, up from 2482)
- [x] New tests added under `tests/` (four, in `tests/naming_tests.rs`)

The four are `a_name_carrying_a_character_windows_forbids_in_a_path_is_refused`, `a_name_whose_segment_is_a_reserved_windows_device_is_refused`, `the_windows_rule_stops_where_the_reserved_list_stops`, and `the_trailing_space_and_period_rules_are_left_to_git`. All exercise the predicate directly, so none needs a Windows filesystem, which is the third acceptance criterion.

## Screenshots / TUI captures

None. No render path changed; the TUI `Ctrl-T` form submits through the same `WorktreeName::freeform` the CLI does, which is the single choke point both entry points share.

## Checklist

- [ ] README updated if CLI surface changed (not needed: the flag is unchanged, only what it accepts)
- [ ] `examples/gwm.toml.example` updated if config schema changed (not needed: no config schema change)

## Linked issues / docs

- Issue: #475
- Related PR: #474 (added `--name` and deferred this)

## Notes for reviewers

Acceptance criteria, in the issue's order:

1. **refused before anything is created** yes, inside `WorktreeName::freeform`, which runs before `worktree::add`.
2. **residual rule set measured rather than assumed** the table above, reproducible from the precondition assertions in the tests.
3. **predicate unit-tested without a Windows filesystem** yes, all four tests call the predicate directly.
4. **docs drop the #474 caveat** yes, EN and FR, plus the `CHANGELOG.md` sentence that said the same thing and the doc comment on `freeform` that documented the exact opposite of what the function now does.

**Left out on purpose.** The issue closes with a wider observation: `worktree::add` creating the branch before the directory is what turns any late failure into an orphaned branch, not just this input set. That is a real fix and a different one. It fires on every late failure rather than on a name, and `add` already tracks `created_branch`, so the rollback has a natural home. Filed as a follow-up rather than folded in here, per the repo's "follow-up issues beat scope creep" rule.


kbrdn1:
Follow-up filed as #487: rolling the branch back when `repo.worktree` fails, which blunts the whole late-failure class rather than this input set.

Issue #479, opened 2026-07-29
[Feature]: rename a free-form worktree — freely again, or into the pattern


kbrdn1:
## Problem

A worktree created with `gwm create --name spike-redis` ([#416](https://github.com/kbrdn1/gwm-cli/issues/416)) cannot be renamed from the TUI at all. The rename form refuses it outright:

```
'spike-redis' is a free-form branch — the edit form rebuilds <type>/#<issue>-<desc>; rename it with git
```

Measured against the current branch, not assumed: `naming::worktree_spec` starts with `branch_parser.parse(branch)?`, a free-form name does not parse, so `enter_edit_worktree` takes the refusal arm and returns.

That leaves two ordinary things impossible:

1. **Renaming it freely again.** A worktree named for a spike outlives the spike; `spike-redis` becomes `spike-valkey`. Today: `git branch -m` plus `git worktree move`, by hand, in two commands that have to agree.
2. **Promoting it into the convention.** The spike turns into real work with an issue number, and there is no way to say "this is now `feat/#42-cache-layer`": the pattern cannot be applied after the fact.

Both are exactly what the rename form exists for, and it is the one shape it refuses.

## Proposed solution

Give the rename modal the same two modes the create form already has since #416, and the same `Ctrl-T` toggle between them.

**Opening.** A branch the pattern reads opens in structured mode as today. A branch it does not read opens in **free-form mode**, with the current branch name prefilled, rather than refusing.

**Submitting.**

| from | to | branch becomes | directory becomes |
|:--|:--|:--|:--|
| free-form | free-form | the name, verbatim | the name, flattened (`/` → `-`) |
| free-form | structured | `branch_pattern` applied to the triple | `path_pattern` applied to the triple |
| structured | free-form | the name, verbatim | the name, flattened |
| structured | structured | unchanged from today | unchanged from today |

The rename worker needs nothing: it already takes a target branch and a target path and does `git branch -m`, the remote rename, and `git worktree move`.

Free-form names are validated by the existing `WorktreeName::freeform` rules: a valid git *branch* name, a single path component under 255 bytes, no `{` / `}`, no leading `-`. Same rules as `gwm create --name`, so a name the rename accepts is a name create would have accepted.

## Decisions this needs, and they are the work

[...]

PR #492, opened 2026-08-03
✨ feat(tui): token-driven create form with pattern-ordered fields

### LINKED ISSUE #418: [Feature]: token-driven create form with live branch/path preview

kbrdn1:
## Description

The TUI create and rename forms presented the canonical `Type` / `Issue` / `Desc`
triple whatever `.gwm.toml` said. They now present the fields the repo's patterns
actually ask for, in the order those patterns write them.

**Read the scope notes below before reviewing.** Two of the issue's three asks had
already shipped, and the issue's scope note declined the design this PR delivers.
Both are addressed explicitly rather than quietly.

Closes #418

## Type of change

- [x] 🐛 Fix (bug fix)

## The defect

Not cosmetic. On a `{type}/{desc}` repo the form demanded an issue number, then
expanded a pattern with nowhere to put it: `BranchSpec::validate_against` refuses
an empty issue, so the value was mandatory *and* discarded. Measured before
touching anything:

```
empty issue -> Some("invalid issue number '' (digits only)")
branch = Ok("feat/thing")     # the 42 went nowhere
dir    = Ok("feat-thing")
```

Three entry points each named a field literally (`reset` chose Type, `toggle_mode`
chose Issue, the rename modal chose Desc), so each was its own way to focus an
input the renderer does not draw, where typing goes nowhere at all.

## Changes

- `naming::editable_segments(patterns)` returns which of `{type}` / `{issue}` /
  `{desc}` a pattern set asks for, in first-appearance order, de-duplicated.
- `CreateForm` carries that field list. `fields` is private and `field` is kept
  inside it: a focused field the renderer does not draw is an input the user can
  neither see nor correct. Rotation walks the list, `push_char` / `pop_char`
  refuse a field outside it, `entry_field` / `last_field` replace the three
  hardcoded choices while preserving what each was for.
- `BranchSpec::new_with_required` / `validate_with_required` validate only the
  segments the patterns carry. Additive: `validate_against` delegates with all
  three, so every existing caller keeps its contract byte for byte.
- `apply_create_form_fields` joins `apply_sidebar_config` at all three of its
  call sites. The Settings one matters on its own: an edit can rewrite the
  patterns, and without it the form kept asking for a token the pattern no longer
  carried until the next launch.
- Both overlays emit their rows from one `form_field_lines`, so they cannot come
  to disagree about which inputs exist.

## The field set includes `base`

`BranchSpec::worktree_path` expands `{type}` / `{issue}` / `{desc}` in
`worktree.base` too (`src/naming.rs:218`), so the union is over **three** patterns,
not two. A segment only `base` carries still names a real directory on disk.

Worth flagging for a follow-up: the guard at the old `app.rs:3954` only ever asked
`branch_pattern` and `path_pattern`, so it was already incomplete on this point.

## Scope notes

**Two of the issue's three asks already shipped.** The live branch/dir preview came
with #217's follow-up and #416 (`pattern_preview`, `edit_target`), and `Ctrl-T`
came with #416 (`modal_keymap.rs`). What was actually left is the field set and the
focus order, which is what this PR is.

**The issue's scope note declined the narrow "show/hide the three" version in favour
of a generic token-driven form, so exotic tokens are supported rather than silently
dropped. That note assumed an open token vocabulary.** The vocabulary is closed:
`expand_placeholders` substitutes seven tokens, and four of them (`{repo}`,
`{home}`, `{repo_path}`, `{repo_parent}`) are resolved from context, never typed.
So the generic form and the narrow one produce an identical field *set* for every
possible pattern, and differ only in *order*. Order is delivered.

The intent behind the note is honoured by a drift guard rather than by machinery
that has nothing to range over: `every_substituted_token_is_either_context_resolved_or_a_form_field`
classifies every token read back out of `expand_placeholders`, so a fourth editable
token added in six months fails a test instead of becoming a field the form
silently never presents.

## Two things reviewers should push back on if they disagree

1. **The type selector moved below the live preview**, where the old layout kept it
   above. The reason is that visual order should be focus order in every pattern,
   including `{desc}/{type}`. This is a visible change on the default pattern too.
2. **The rename modal's refusal guard is deleted**, and the test that pinned its
   message now asserts the reverse. Its premise was a form hardcoded to the triple
   ("the form cannot be submitted without it"), which stopped being true.

## Deliberately not in scope

- **The preview still shows the directory name, not the absolute path.** The path
  prefix only varies when `base` itself carries a token, and adding a third preview
  row is a UI decision rather than a correctness one.
- **`gwm create feat 42 thing` still demands a discarded `42` on a no-issue
  pattern.** Same defect, other surface. Changing positional arity is breaking and
  lands in frozen-contract territory, so it belongs in its own issue.

## Tests

- [x] `cargo test` passes locally (88 test binaries, exit 0), run under a CI-like
      minimal PATH: `PATH="$(dirname "$(command -v cargo)"):/usr/bin:/bin" cargo test`
- [x] `gwm doctor` clean

Test diff, by file:

- `tests/naming_tests.rs`: pattern order, the `base` union, de-duplication, and the
  drift guard.
- `tests/tui_state_create_form_tests.rs`: field set per pattern shape, rotation that
  cannot land off-list, every entry point landing on a presented field, typing into
  an omitted field, and the degenerate all-literal pattern.
- `tests/tui_app_tests.rs`: the end-to-end submit without an issue number, the
  regression guard that a segment the pattern *does* carry is still required, and
  the reversed rename-modal test.
- `tests/tui_modal_render_tests.rs`: what is actually drawn, for both overlays.

The issue sanctions the rewrite of the state test file: "Expect the test file to be
substantially rewritten rather than extended." Two rotation tests were rewritten in
place and one `tui_app_tests.rs` test was reversed; nothing was deleted for
convenience.

## Screenshots / TUI captures

Not captured. This is a TUI change and wants a local install before merge.

## Checklist

## Linked issues / docs

- Issue: #418
- Sequence: #415, #416, #417, #479 all merged; this is the last step of the naming
  sequence.

## Notes for reviewers

The interesting review surface is the two pushback points above, and whether the
`base` union should also be applied retroactively to the parse-side guard that only
asked two patterns.


kbrdn1:
Follow-up after review: **two blocking defects found on the key-handler path**, which
none of the state or render tests reach. Both fixed in `eb324ce`, both with a test
that goes through `handle_create_key` / `submit_create` rather than around them.

1. **`submit_create` composed its own `BranchSpec::new_with_types`.** Relaxing
   validation in `worktree_name_from_form` fixed the *rename* path and left *create*,
   the path this issue is actually about, still failing with `invalid issue number ''`.
   The headline test passed only because `edit_target` is the rename composer. Both now
   go through the one function. Two composers of one value drift, and this is the second
   time on this form: the previews did it in #476.

2. **Enter submitted only from `Field::Desc | Field::Name`**, those being the last
   fields of the two hardcoded modes. Making `{desc}` optional turned `{type}/#{issue}`
   into a form that **could not be submitted at all**, since Enter only ever rotated.
   A bug this PR would have introduced, not a pre-existing one. It now gates on the last
   field in pattern order.

Two smaller things came with them:

- The status line said "enter on desc: submit" regardless of the pattern. It now reads
  the same `last_field` the key handler gates on, so hint and behaviour cannot disagree.
- `refuse_unwritable_segment_change` open-coded the same "does any pattern write this"
  test that `editable_segments` now answers. Routed through the shared source. Worth
  noting for review: that guard is **unreachable in structured mode** as a result, since
  the form presents no field for a segment nothing writes. Kept as a backstop rather than
  deleted, because three earlier versions of it were wrong in ways the field set does not
  speak to.

**Filed rather than fixed here: #493.** The hint row above the form still advertises
`↑/↓ type` on a pattern with no `{type}`, and `field` on a single-field pattern. Same
class as the field set, and this repo took a Codex P2 for exactly that shape on #485.
It is not in this PR because `HintContext::hints()` returns `&'static [Hint]` from a
table keyed on the context alone, so making it pattern-aware changes a surface shared
by every modal in the TUI.

Suite re-run green after the fixes: 88 test binaries, exit 0, under the CI-like minimal
PATH.

kbrdn1:
## Review loop: clean on pass 7 of 8

`P0P1P2_COUNT=0`, verified against the real report rather than an empty parse
("Aucun défaut bloquant ou régression observable n'a été identifié").

| pass | findings | worst | what it was |
|---|---|---|---|
| 1 | 2 | P2 | trailing period per segment; the opening instruction naming `desc` |
| 2 | 2 | P2 | a base-only type silently defaulted; hint rows advertising dead keys |
| 3 | 1 | P2 | field set derived from the raw template |
| 4 | 1 | P2 | my bound on pass 3 was false |
| 5 | 2 | P2 | strings naming absent fields, in two more places |
| 6 | 1 | **P1** | a hidden segment blocking every structured rename |
| 7 | 0 | - | clean |

Six of the eight findings are worth reading, because four of them were in code
or claims I had just written.

**The P1 disproved a comment I had added two passes earlier.** I wrote that
`refuse_unwritable_segment_change` had become unreachable in structured mode,
reasoning that the form presents no field for a segment nothing writes, so its
value cannot differ from what the form opened with. Wrong: the buffer behind a
hidden field keeps its *default*, and `type_index` defaults to the first
configured type. So on a pattern set omitting `{type}` the guard compared `feat`
against the empty parsed type, called it a change, and refused every structured
rename with `branch_pattern has no {type} to write`, with no field on screen to
clear it.

**Pass 2 caught the most expensive one, and it was my own fix from pass 2 being
wrong in the other direction.** I had deleted the guard that refused to open the
rename form on an unrecovered segment, on the grounds that a segment no pattern
carries is not a dead end. True, but `worktree_spec` reads the branch and the
directory name and never parses `base`, so a segment carried only by `base` comes
back empty *while its value sits on disk*. Under `base = ".../wt/{type}"` a
worktree at `.../wt/fix/my-desc` would have opened showing the first configured
type, and submitting without touching it would have moved the worktree to
`.../wt/feat/`. "Not recovered" is not "not there". The rule that ships is the
conjunction: refuse a segment some pattern writes that the parse did not recover.

**Pass 4 broke the justification I gave on pass 3.** I had argued the raw-template
scan was bounded by a failure that already ships, since #417's compiler refuses
patterns whose expansion carries a token and `branch_pattern_warning` reports
them. That bound is false: the warning is only ever called with
`worktree.branch_pattern`, so the same repo name reaches the form through
`path_pattern` or `base` with no diagnostic at all. The scan now mirrors the
formatter. Root cause filed as #494.

**Pass 5 was the third finding of one class**, so it stopped patching strings.
Grepping the surface found five sites naming a field the patterns may not present,
not the two reported, three of them in the help overlay. The guard is now the
property: on a repo whose patterns omit a segment, nothing the create or rename
surface renders may name it. One assertion, both modes, covering the modal body,
its footer and the statusbar.

⚠️ **That guard was toothless on its first draft.** It called
`CreateForm::toggle_mode` directly, but the status line is written by the key
handler, so the string under test was never produced: it passed with the defect
deliberately restored. It goes through `handle_create_key` now, verified red with
the old string and green without. Same shape as the two blockers from the first
review round, which were also invisible to state and render tests.

**#493 closed rather than kept.** I filed it to defer the hint row, on the
assumption that making it pattern-aware meant threading form state through
`modal_hint_for_context`'s eight call sites. Only two of those eight render a
create or rename footer, so `resolve_with_fields` carries the field set and the
seven other modals are untouched. It also mattered that the inaccuracy is
introduced by this PR rather than pre-existing.

## Still open, deliberately

- **#494**, `expand_placeholders` substituting what its own expansions produce.
  Reaches the CLI, hooks and path resolution, so it wants its own change.
- **A segment carried only by `base` is not recovered from disk**, so the rename
  form refuses rather than prefilling. Recovering it means compiling a parser for
  `base`. Stated in the code, not silently accepted.
- **`gwm create feat 42 thing` still demands a discarded `42`** on a no-issue
  pattern. Same defect, other surface; positional arity is frozen.

CI green on 8 of 9 checks with `test (windows-latest)` still running.

This is a TUI change, so it wants a local install before merge.

PR #495, opened 2026-08-03
🐛 fix(config): substitute worktree placeholders in a single pass

### LINKED ISSUE #494: [Fix]: expand_placeholders substitutes what its own expansions produce
## Problem

`config::expand_placeholders` substitutes `{home}` and `{repo}` first, then substitutes
`{type}` / `{issue}` / `{desc}` over the result. So it substitutes text that its own
expansions produced. Measured:

```
repo    = "api-{type}"
pattern = "{repo}/{desc}"
expand_placeholders(pattern, repo, Some("fix"), Some("42"), Some("foo"), None)
  -> Ok("api-fix/foo")
```

The branch carries a type the template never mentions, so anything that reasons about
the pattern by reading it is wrong on this repo. Found by Codex review on #492.

## Corrections to this issue's first draft

Both are mine, and both change what the fix has to do.

**`lifecycle` is not affected.** I wrote that this reaches hook placeholders. It does
not: `lifecycle::expand` has been single-pass since the security patch (`97e4e09`),
walking the template once and never re-scanning what it wrote.
`tests/lifecycle_tests.rs` already pins it with a branch literally named
`spike-{issue}`. That expander is the reference implementation for this one.

**#418's workaround becomes actively wrong.** `naming::editable_segments` now resolves
`{repo}` / `{home}` before scanning, so the create form derives the fields the
multi-pass formatter will actually fill. Once the formatter stops re-substituting, that
expansion makes the form ask for a field the pattern does not write. Fixing the root
means reverting it, which is the point of the change rather than a side effect.

**The bound I first gave was also wrong**, recorded because it looked solid: I argued
`BranchParser::compile` refuses this class and `branch_pattern_warning` reports it. The
warning is only ever called with `worktree.branch_pattern` (`doctor.rs`,
`config_cli.rs`), so the same repo name reaches `path_pattern` and `base` with no
diagnostic at all.

## Proposal

Make the substitution single-pass: walk the template once, replace each recognised token
with its value, and never re-examine what was written. Expansions become opaque text,
which is what every consumer already assumes.

Two behaviours the rewrite must preserve rather than tidy away:

- A token with no value is left **literal**, not emptied. `BranchSpec::branch_name`
  passes no `repo_path` precisely so `{repo_path}` / `{repo_parent}` survive verbatim
  into the branch name, and `BranchParser::compile` mirrors that.
- The asymmetry when `repo_path` is `Some(p)` but `p.parent()` is `None`: `{repo_path}`
  resolves, `{repo_parent}` stays literal. Today that falls out of the nesting; it has
  to be stated explicitly in a token lookup.

This is a widening, not a restriction: `BranchParser::compile` currently refuses
`{repo}/{desc}` on a repo whose name contains a brace, and `mirrors_formatter` will
accept it once the two agree. Its error message loses the clause blaming multi-pass
substitution, but the check keeps its job (the `~` exclusion means other divergences
remain possible).

## Tests

- `tests/config_tests.rs`: a repo name **and** a `$HOME` containing each of the three
  tokens; assert the expansion leaves them literal. Verified by restoring the defect and
  confirming red, since a fixture whose values contain no token passes either way.
- `tests/config_tests.rs`: the `repo_path` / `repo_parent` asymmetry above.
- `tests/naming_tests.rs`: `a_token_produced_by_expanding_the_repo_name_is_still_a_field`
  inverts, since the pattern then genuinely writes no type.
- `tests/naming_tests.rs`: `{repo}/{desc}` on such a repo now compiles and round-trips.

## Also stale, found while measuring

`WorktreeName::freeform` rejects `{` and `}` citing `lifecycle::expand_placeholders`
re-substituting a name that contains a token. That behaviour is gone. The rejection is
worth keeping as a defensive rule, but its stated cause is false, and a rejection
justified by a false cause is how the next person removes it for the wrong reason.
Whether to loosen it is a naming-surface decision, not part of this fix.

kbrdn1:
## Description

`config::expand_placeholders` chained `str::replace` calls, so each one re-scanned
what the previous had written. `{home}` and `{repo}` go first, which means a repo
whose own directory is named `api-{type}` made `{repo}/{desc}` write `api-fix/foo`:
a branch type the template never mentions.

**An expansion is a value, not more template.** It now substitutes in one pass.

Closes #494

## Type of change

- [x] 🐛 Fix (bug fix)

## Why it matters beyond the pathological repo name

Two things reason about a pattern by reading it, and both were wrong there:

- `naming::editable_segments` derives the create form's field set (#418).
- `BranchParser::compile` derives the branch parser, and could not mirror a
  formatter that rewrote its own output. It refused the whole class rather than
  build a parser recognising none of the branches the pattern creates.

So this is a **widening**: `{repo}/#{issue}-{desc}` on a repo named `{type}`, and
`{{repo}}/#{issue}-{desc}`, both compile and round-trip now. The test that pinned
their refusal asserts that instead.

## Three things preserved rather than tidied away

1. **A token with no value stays literal**, not emptied. `BranchSpec::branch_name`
   passes no `repo_path` precisely so `{repo_path}` / `{repo_parent}` survive
   verbatim into the branch name, and the compiler mirrors that.
2. **The `repo_path` / `repo_parent` asymmetry.** With `repo_path` set but its
   parent `None` (`/` is the case), `{repo_path}` resolves and `{repo_parent}`
   does not. The old nested `if let` fell into that; a token lookup has to state
   it. Pinned by its own test.
3. **The token starts at the last `{` before the closing brace**, not the first.
   `str::replace` matched `{type}` inside `{{type}` and wrote `{feat`, and #417
   built the compiler to mirror that. My first draft scanned from the first brace,
   which left `{{type}` literal instead: a pattern that worked in 1.5.0 quietly
   changing meaning, which is a different bug from the one being fixed. Caught by
   `the_compiler_finds_placeholders_where_the_formatter_substitutes_them`.

## Corrections to the issue, made before coding

Both were mine, and both changed what the fix had to do. The issue body is updated.

- **`lifecycle` is not affected.** I had written that this reaches hook
  placeholders. `lifecycle::expand` has been single-pass since the hook-injection
  patch, walking the template once and never re-scanning what it wrote, and
  `tests/lifecycle_tests.rs` already pins it with a branch named `spike-{issue}`.
  That expander is the reference implementation for this one.
- **#418's workaround becomes actively wrong.** `editable_segments` resolved
  `{repo}` / `{home}` before scanning so the form would ask for the fields the
  multi-pass formatter was going to fill. Once expansions are values, resolving
  there invents a field for a token nothing substitutes. Reverting it is the point
  of the change rather than a side effect, so it is in its own commit.

## Also stale, found while measuring

`WorktreeName::freeform` rejects `{` and `}` citing
`lifecycle::expand_placeholders` rewriting a name that contains a token. That
cause was already gone before this PR. The rule is **kept** as defensive, since
loosening what a free-form name may contain is a naming-surface decision and not
part of this fix, but its comment now says the reason has expired. A rejection
justified by a false cause is how the next reader removes it for the wrong reason.

## Tests

- [x] `cargo test` passes locally (88 test binaries, exit 0), under a CI-like
      minimal PATH: `PATH="$(dirname "$(command -v cargo)"):/usr/bin:/bin" cargo test`
- [x] `gwm doctor` clean

**Verified by restoring the defect and confirming red.** A fixture whose values
contain no token passes on the multi-pass code too, so it would otherwise prove
nothing. Both new `config_tests` go red with the replace chain put back.

New coverage:

- `tests/config_tests.rs`: a token arriving through the repo name (all five
  tokens, so no single ordering of the old chain passes by luck), a token
  arriving through `$HOME`, the no-value-stays-literal contract including an
  unknown token and an unbalanced brace, and the `repo_path` / `repo_parent`
  asymmetry.
- `tests/naming_tests.rs`: the two previously-refused patterns now round-trip,
  and the field set reads the pattern as written.

⚠️ Two existing tests in `config_tests.rs` read `{home}`, so the new `$HOME` test
would have raced them. The binary gains the process-global `env_lock` the other
env-touching suites already use, and the three tests take it.

⚠️ The two source-scanning guards in `naming_tests.rs` extracted tokens by
anchoring on `("{`, which found nothing once the replace chain became a `match`.
They passed while classifying an empty set; the `found.len() >= 5` floor is what
caught it, which is why that assertion is written as a floor rather than a
comment.

## Checklist

## Linked issues / docs

- Issue: #494
- Found by Codex review on #492, which is where the symptom was worked around.

## Notes for reviewers

The interesting question is whether the last-brace rule is the right tokenisation
or whether `{{` should mean an escaped brace. This PR keeps 1.5.0's meaning
deliberately; changing it is a separate decision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -173,6 +173,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `worktree` patterns are expanded in a single pass, so an expansion is a value
+  rather than more template. `expand_placeholders` chained `str::replace` calls
+  and each one re-scanned what the previous had written, with `{home}` and
+  `{repo}` going first: on a repo whose own directory is named `api-{type}`,
+  `branch_pattern = "{repo}/{desc}"` wrote `api-fix/foo` and carried a branch
+  type the pattern never mentions.
+
+  Anything that reasons about a pattern by reading it was wrong there, and two
+  things do. The create form derived its fields from the wrong set, and the
+  branch parser could not mirror a formatter that rewrote its own output, so it
+  refused the whole class rather than build a parser recognising none of the
+  branches the pattern creates. Those patterns now compile and round-trip.
+
+  Two behaviours are preserved rather than tidied away: a token with no value
+  stays literal instead of collapsing to empty, which is what lets
+  `{repo_path}` and `{repo_parent}` survive into a branch name; and the token
+  starts at the last `{` before the closing brace, so `{{type}` keeps writing
+  `{feat` as it did in 1.5.0.
+
+  Hook placeholders were already single-pass and are unaffected.
+  ([#494](https://github.com/kbrdn1/gwm-cli/issues/494))
+
 - `gwm pr`, `gwm commit-prefix` and `gwm bootstrap` read the branch of the
   worktree they were pointed at, instead of the main checkout's.
   `worktree::discover_repo` walks back to the main working directory when it
--- a/src/config.rs
+++ b/src/config.rs
@@ -2080,22 +2080,63 @@ pub fn expand_placeholders(
     .ok_or_else(|| GwmError::Config("cannot resolve $HOME".into()))?
     .to_string_lossy()
     .to_string();
-  let mut out = template.replace("{home}", &home).replace("{repo}", repo);
-  if let Some(t) = type_ {
-    out = out.replace("{type}", t);
-  }
-  if let Some(i) = issue {
-    out = out.replace("{issue}", i);
-  }
-  if let Some(d) = desc {
-    out = out.replace("{desc}", d);
-  }
-  if let Some(p) = repo_path {
-    out = out.replace("{repo_path}", &p.to_string_lossy());
-    if let Some(parent) = p.parent() {
-      out = out.replace("{repo_parent}", &parent.to_string_lossy());
+
+  // Resolved value for one `{token}`, or `None` to leave it verbatim.
+  //
+  // The `None` arm carries real weight: `BranchSpec::branch_name` passes no
+  // `repo_path` precisely so the disk-path tokens survive into the branch
+  // name, and `BranchParser::compile` mirrors that by treating them as
+  // literals. Note the asymmetry when a `repo_path` has no parent — `/` is the
+  // case — where `{repo_path}` resolves and `{repo_parent}` does not: the old
+  // nested `if let` fell into that by accident, so it has to be said here.
+  let value_for = |token: &str| -> Option<String> {
+    match token {
+      "{home}" => Some(home.clone()),
+      "{repo}" => Some(repo.to_string()),
+      "{type}" => type_.map(str::to_string),
+      "{issue}" => issue.map(str::to_string),
+      "{desc}" => desc.map(str::to_string),
+      "{repo_path}" => repo_path.map(|p| p.to_string_lossy().into_owned()),
+      "{repo_parent}" => repo_path
+        .and_then(|p| p.parent())
+        .map(|p| p.to_string_lossy().into_owned()),
+      _ => None,
     }
+  };
+
+  // One pass, never re-examining what was written (issue #494). Chained
+  // `str::replace` calls re-scan the previous call's output, so a value that
+  // itself contains a token got rewritten from the inside: on a repo named
+  // `api-{type}`, `{repo}/{desc}` wrote `api-fix/foo` and carried a type the
+  // template never mentions. **An expansion is a value, not more template** —
+  // nothing downstream can reason about a pattern by reading it otherwise, and
+  // `naming::editable_segments` and `BranchParser::compile` both do.
+  //
+  // Same property `lifecycle::expand` has had since the hook-injection patch,
+  // and for the same reason one layer down.
+  let mut out = String::with_capacity(template.len());
+  let mut rest = template;
+  while let Some(open) = rest.find('{') {
+    let Some(close) = rest[open..].find('}').map(|at| open + at) else {
+      // Unbalanced `{` — no token to close, so the remainder is literal.
+      break;
+    };
+    // The token starts at the **last** `{` before this `}`, not the first.
+    // `str::replace` matched `{type}` inside `{{type}` and wrote `{feat`, and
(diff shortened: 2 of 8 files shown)


kbrdn1:
## Review loop clean on pass 2 of 8, CI green on all ten checks

| pass | findings | worst | what it was |
|---|---|---|---|
| 1 | 1 | **P1** | the `$HOME` test cannot work on Windows |
| 2 | 0 | - | clean |

**The P1 was confirmed by CI before the loop reported it**, on the very commit that
introduced it. `dirs 6` resolves the home directory through `SHGetKnownFolderPath` on
Windows and never reads `$HOME`, so swapping the variable there changes nothing and the
assertion fails every run. That is CLAUDE.md's rule about pre-validating
environment-dependent tests being paid for rather than followed. The test is now
`#[cfg(unix)]`, and the property it covers stays portable: `{home}` and `{repo}` are two
arms of one `value_for` match, and the repo-name test exercises all five tokens arriving
through a value.

## One correction to the PR, after checking a claim I had not checked

The commit `84d1787` walks back something I wrote earlier in this branch.

I had rewritten `WorktreeName::freeform`'s comment to say its rejection of `{` and `}`
now stands without the bug that motivated it, since both expanders are single-pass.
**I had checked two expanders and there are three.** `launcher::expand` still chains
`str::replace` over its own output, and it substitutes the worktree path *before*
`{base}` / `{head}`, so a brace reaching that path is re-substituted inside an already
shell-quoted value.

`kebab` strips braces out of a structured description and `freeform` refuses them, which
leaves that check as the only thing keeping a brace out of a chosen name that becomes a
directory. Measured rather than read: `kebab("a{base}b")` is `a-base-b`,
`freeform("a{base}b")` is refused, and `worktree.base` passes one through verbatim
(`/tmp/{base}/feat-x`), which is config rather than a name and so a separate surface.

So the rule is not historical, and the comment says so now. Calling a live guard
historical is worse than leaving a stale reason on it.

## Scope, stated precisely

This PR fixes `config::expand_placeholders`. It does **not** touch `launcher::expand`,
which has its own token set (`{path}`, `{base}`, `{head}`, `{diff}`) and its own
quoting rules. `lifecycle::expand` was already single-pass and is unaffected.
`doctor.rs` calls `config::expand_placeholders` directly and is covered.

So #494 removes the multi-pass shape from the pattern expander, not from the codebase.

PR #496, opened 2026-08-03
🔒 fix(cli): neutralise the config strings every ungated command echoes

### LINKED ISSUE #473: [Fix]: config values are echoed raw, so an unvetted .gwm.toml can drive the terminal
## Problem

`gwm config list` and `gwm config get` print config values verbatim:

```rust
// src/config_cli.rs
println!("{} = {}", key, value);
```

Config values come from the repo's `.gwm.toml`, and the `gwm config` read commands do **not** go through the [TOFU trust gate](/configuration/trust-ledger) — inspecting an unfamiliar repo's config is meant to be a safe thing to do. A `.gwm.toml` carrying control bytes in any string value therefore gets a terminal escape channel out of a read-only command: an OSC 52 clipboard write, a window-title rewrite, cursor manipulation to hide subsequent output.

This is pre-existing on `dev`, not a regression.

## How it surfaced

Found by a Codex review on #472 (#415), which added a *new* echo path for `worktree.branch_pattern` in `gwm doctor` / `gwm config validate`. That new path was fixed in the same PR — both values now go through a control-character replacement before being formatted. The pre-existing `config list` / `config get` surface was deliberately left out of that PR rather than widening its scope, per CLAUDE.md's "follow-up issues beat scope creep".

## Proposal

Apply the same idiom the codebase already uses in three places
(`tui/wt_tree.rs::sanitize_name`, `tui/ui.rs`, `agent_sessions.rs`): replace
control characters rather than strip them, so the value stays recognisable and
its length is not silently altered.

```rust
fn sanitise_for_terminal(s: &str) -> String {
  s.chars().map(|c| if c.is_control() { '?' } else { c }).collect()
}
```

Call sites to audit — this should be a sweep, not a one-liner, since the same reasoning applies anywhere a config-supplied string reaches stdout/stderr outside the TUI (which already sanitises):

- `config_cli::list` — the `{} = {}` loop
- `config_cli::get` / `set` — `format_get_value` / `format_list_value` output
- any other `println!` of a `.gwm.toml`-sourced value

## Tests

`tests/config_cli_tests.rs`: a `.gwm.toml` whose value carries `\u{1b}]52;c;…\u{7}` produces stdout with no control characters, and the value is still recognisable in the output. Mirrors `naming_tests::control_characters_in_the_pattern_never_reach_the_terminal` added by #472.

## Related

- #472 / #415 — the doctor / config-validate echo path, already fixed

kbrdn1:
## Description

A `.gwm.toml` is data from a repo nobody has vetted, and the commands that read
it back skip the TOFU trust gate on purpose: inspecting an unfamiliar repo
before trusting it is meant to be the safe move. Echoing a config-supplied
string verbatim handed that file a terminal escape channel out of a read-only
command.

Closes #473

## Type of change

- [x] 🐛 Fix (bug fix)

## Changes

Measured against a hostile `.gwm.toml` **before** the fix, each of these put a
raw ESC on the terminal:

| command | site |
|---|---|
| `gwm config get` | the string arm printed the value bare |
| `gwm config list` | the **keys** (`aliases`, `forge_hosts`, `exec`/`clean` profiles, `tui.keys` are maps, so their keys are attacker-chosen) |
| `gwm types` | `description` from `[[branch_types]]`, shortcodes from `[gitmoji]` |
| `gwm aliases list` | repo and user alias names and expansions |
| `gwm doctor` | check details that quote a config value |
| `gwm commit-prefix` | assembled from `[gitmoji]` |
| the TOFU prompt | the bootstrap surface, and the raw body on `show` |
| **any command at all** | `toml`'s parse error quotes the offending source line |

The last two are the ones that matter.

The bootstrap summary renders directly above
`Trust this .gwm.toml? [y/N/show]:`, so a cursor-up-then-erase-line pair in a
`[[bootstrap.command]]` name walks the cursor up and deletes the row naming the
shell it is asking permission to run. The summary whose job is to reveal the
command can be made to hide it.

The parse-error leg needs no echo command at all: running `gwm list` inside the
repo is enough.

Sanitised at each **sink** rather than at each producer. That is what makes the
doctor leg cover checks nobody has written yet, and what makes the two
`eprintln!` in `main` cover every error of every subcommand.

Two variants, because the sinks have two shapes:

- `sanitise_for_terminal`, for rows. Replaces every control character, newlines
  included, so a value cannot forge an extra row in a report read as a list of
  facts.
- `sanitise_block_for_terminal`, for documents. Keeps the line break and the
  tab, which carry the layout, and replaces the rest including the carriage
  return, which would let a value overwrite the line it was printed on.

Sanitising the `show` body is not "breaking raw": an escape sequence buried in
the body defeats the inspection that view exists to provide.

## Where the issue was wrong

#473 names `println!("{} = {}", key, value)` in `config list` as the problem.
The **value** there was never the hole: `format_list_value` renders strings
with `{:?}`, and Rust's `Debug` for `str` escapes control characters. Verified
by running the binary, not by reading the impl. The **key** beside it was raw,
which is the leg that actually needed fixing.

Rather than change that rendering, a test now pins the escaping. The protection
is incidental to a formatting choice, so "simplifying" `{:?}` to `{}` would
reopen the channel silently.

## Deliberately out of scope

- **`labels` / `milestones` diff rows.** Their printers are only reachable
  after a forge round-trip, so there is no test seam without mocking `gh`, and
  this repo does not ship untested behaviour. Label names are already
  control-char-validated at load (`labels::validate_label_name`); milestone
  titles are not, which is worth a follow-up.
- **The bootstrap report** (`lifecycle.rs`, `cli.rs`) runs only once the gate
  has passed or `--allow-bootstrap` was given, so the user has already
  consented to that file.
- **`exec` / `clean` profiles** *execute* config-supplied commands. That is a
  different and larger threat than terminal escapes, and not this PR's.

## Tests

- [x] `cargo test` passes locally: 89 targets, exit 0, under a CI-like minimal
      PATH (`PATH="$(dirname "$(command -v cargo)"):/usr/bin:/bin"`)
- [x] New tests added under `tests/terminal_escape_tests.rs`

Five tests, written red before the fix and pinned at the sink rather than at
the call site, so the guarantee survives a new check / row / field upstream:

- `no_ungated_command_replays_a_control_byte_from_the_repo_config`, a table
  over the seven ungated commands, stdout **and** stderr
- `a_neutralised_value_is_still_recognisable_in_the_output`
- `config_list_keeps_escaping_string_values`, the incidental `Debug` pin
- `a_config_parse_error_does_not_replay_the_files_control_bytes`
- `the_pre_trust_bootstrap_summary_neutralises_control_bytes`

One thing the fixture documents: the payload has to travel as TOML's own `\u`
escape, because the parser **refuses** a raw control byte inside a basic
string. A raw byte never becomes a value at all, it produces a parse error
instead, which is exactly why the parse-error leg exists.

## Checklist

- [ ] README updated if CLI surface changed: no surface change, output only
- [ ] `examples/gwm.toml.example` updated if config schema changed: no schema change
- [x] No `println!` in TUI render code. TUI untouched; it already sanitises.

## Linked issues / docs

- Issue: #473
- Related PR: #472 / #415, which fixed the `doctor` / `config validate`
  `branch_pattern` echo and introduced `sanitise_for_terminal`, reused here

## Notes for reviewers

`a6e8eeb` is a pure refactor with no behaviour change: it turns
`print_bootstrap_summary` into `bootstrap_summary_lines` returning
`Vec<String>` so the highest-severity site can be asserted on without driving a
PTY. Otherwise the important half of this PR would have shipped untested.

Widths in `types` and `aliases list` are now measured on the neutralised
strings, because a replaced C1 control character is two bytes narrower than the
one it replaced, so measuring the raw string would skew the columns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -173,6 +173,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Config values no longer reach the terminal with their control bytes intact.
+  A `.gwm.toml` comes from a repo nobody has vetted, and the commands that read
+  it back skip the trust gate on purpose, because inspecting an unfamiliar repo
+  before trusting it is meant to be the safe move. Echoing a value verbatim
+  handed that file a terminal escape channel out of a read-only command: an
+  OSC 52 clipboard write, a window-title rewrite, cursor moves that erase the
+  line above.
+
+  `gwm config get`, `config list` (its keys, which are attacker-chosen wherever
+  the schema is a map), `types`, `aliases list`, `doctor`, `commit-prefix` and
+  the TOFU prompt are all covered. So is `toml`'s parse error, which quotes the
+  offending source line verbatim and therefore needed no echo command at all —
+  running `gwm list` inside the repo was enough.
+
+  The one that matters most is the prompt: its bootstrap summary renders
+  directly above `Trust this .gwm.toml? [y/N/show]:`, so a cursor-up-and-erase
+  in a `[[bootstrap.command]]` name could delete the row naming the shell it
+  was asking permission to run.
+
+  Values are replaced rather than stripped, so what is left stays recognisable
+  and no length changes silently. Output that is meant to span rows — a parse
+  diagnostic, the raw body shown by `show` — keeps its line breaks and loses
+  everything else.
+
 - `worktree` patterns are expanded in a single pass, so an expansion is a value
   rather than more template. `expand_placeholders` chained `str::replace` calls
   and each one re-scanned what the previous had written, with `{home}` and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -182,10 +182,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   line above.
 
   `gwm config get`, `config list` (its keys, which are attacker-chosen wherever
-  the schema is a map), `types`, `aliases list`, `doctor`, `commit-prefix` and
-  the TOFU prompt are all covered. So is `toml`'s parse error, which quotes the
-  offending source line verbatim and therefore needed no echo command at all —
-  running `gwm list` inside the repo was enough.
+  the schema is a map), `types`, `aliases list`, `doctor`, `commit-prefix`,
+  `trust list` / `show` / `add` / `revoke`, the milestone and label diff rows,
+  and the TOFU prompt are all covered. So is every error message, which needed
+  no echo command at all: `toml`'s parse error quotes the offending source line
+  verbatim, so running `gwm list` inside the repo was enough.
 
   The one that matters most is the prompt: its bootstrap summary renders
   directly above `Trust this .gwm.toml? [y/N/show]:`, so a cursor-up-and-erase
@@ -193,9 +194,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   was asking permission to run.
 
   Values are replaced rather than stripped, so what is left stays recognisable
-  and no length changes silently. Output that is meant to span rows — a parse
-  diagnostic, the raw body shown by `show` — keeps its line breaks and loses
-  everything else.
+  and no length changes silently. Errors keep their line breaks, because a
+  parse diagnostic points at the broken column across several rows, but every
+  line after the first now sits under gwm's own margin. That is what stops a
+  value carrying a newline from printing a second line at column zero that
+  reads like a statement from gwm rather than a quote from the repo.
+
+  One value is refused rather than neutralised: an `[aliases]` expansion
+  becomes argv before the CLI parser runs, and that parser prints its own
+  errors without passing through gwm's output path, so a control character
+  there is rejected when the config loads.
 
 - `worktree` patterns are expanded in a single pass, so an expansion is a value
   rather than more template. `expand_placeholders` chained `str::replace` calls
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to this project will be documented here.
 
 This file tracks the **in-progress** release only. Past releases live under
-[`changelogs/`](changelogs/) — one Markdown file per SemVer version.
+[`changelogs/`](changelogs/), one Markdown file per SemVer version.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- RELEASE NOTE: before publishing, rewrite BOTH sections below (Added and
      Fixed) in the advisory's own wording and reference the GHSA id. The full
-     write-up — impact, threat model, reproduction — lives in the draft
+     write-up (impact, threat model, reproduction) lives in the draft
      advisory, not in this file, until it is published. -->
 
 ### Added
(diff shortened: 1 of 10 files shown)


kbrdn1:
## Review loop: stopped on convergence, not on a clean pass

Six passes of the local Codex loop. Stopping here on `exit_on_stagnation`, not
on `P0P1P2_COUNT=0`, and the distinction matters: the last count I saw from
the real check output was **2**, both P2, both fixed in `72b1067` but not
re-verified by a seventh pass.

| pass | findings | P0 | P1 | P2 | nature |
|---|---|---|---|---|---|
| 1 | 2 | 0 | 2 | 0 | coverage gaps in this PR's own new code |
| 2 | 2 | 0 | 2 | 0 | one real hole (`trust list`), one refuted (clap) |
| 3 | 2 | 0 | 1 | 1 | one class fix (forged line), one real hole (forge diff rows) |
| 4 | 1 | 0 | 1 | 0 | a defect in pass 3's fix |
| 5 | 1 | 0 | 1 | 0 | a defect in pass 2's test |
| 6 | 2 | 0 | 0 | 2 | two readability regressions this PR introduced |

**Where it turned.** Passes 1 to 3 found holes in the product: commands
echoing a config value raw. From pass 4 on, every finding pointed at a line
written to fix an earlier pass. Three consecutive, where the documented stop
condition is two.

**Severity trajectory.** Pass 1's worst was a terminal escape channel out of
`gwm config get`. Pass 6's worst was a stray `?` at the end of each line of
`trust show` on Windows. Both worth fixing; they are not the same kind of
problem, and the gap between them is the argument for stopping.

### What each pass changed

- **1** `commit-prefix` and `trust show` had a sanitiser each that no test
  exercised. Deleting either left the suite green.
- **2** `trust list` reloads the ledger and prints a *decoded* origin, so
  cleaning only the raw file body protected an already-escaped representation.
  Fixed there plus `revoke` / `add` / the post-approval line. The clap leg was
  **refuted**: measured on three shapes, clap strips the ASCII controls from
  the token it quotes and the expander splits on whitespace, so a `\n` becomes
  a second token rather than a forged line.
- **3** A config value carrying `\n` could forge a line reading
  `error: everything is fine` in gwm's own voice. Third finding of that class,
  so the fix moved to the sink instead of another call site. Also made the
  milestone and label diff rows values, which the first cut had wrongly called
  untestable.
- **4** Refuted pass 3's approach on its own terms. Classifying whole messages
  cannot work: `toml` emits generated layout and a repo-written key in one
  string. Replaced by owning the left margin, which deleted the
  `ConfigDiagnostic` variant it replaced. Net simpler.
- **5** Pass 2's refutation was right about clap and wrong about the range.
  Clap leaves the C1 codes: `\u{9b}` (CSI) and `\u{85}` (NEL) came out intact.
  The expansion is now refused at load rather than cleaned at a sink we do not
  own.
- **6** `gwm doctor` rendered `?  |?1 | [worktree?` for a broken config, on the
  command whose job is recovery. And a CRLF pair was read as an escape, so
  Windows saw a `?` at every line end.

### Findings not fixed

None outstanding. Two refuted with measurements rather than argument (clap's
handling of ASCII controls; `trust show` protecting an already-escaped body),
and both refutations were themselves later corrected where they were too
broad.

### Three of these tests passed against the defect they claimed to cover

Worth naming, because each failed differently and a green security test is
worse than no test:

1. The assertion did not match real output. `l.trim() == "error: ..."` never
   fires, because the forged row runs on into the rest of the real message.
2. The fixture never reached the path. `str::escape_default` emits Rust's
   `\u{9b}`; TOML rejects the braces, so the config did not parse, alias
   loading failed for the wrong reason, and the assertion ran against output
   the payload had never touched.
3. The command failed before its printer. `commit-prefix` was added to a
   shared matrix whose `branch_pattern` starts with an ESC, which no git-legal
   branch can match, so `main`'s error sink was doing the cleaning.

Every guard in this PR was subsequently run with its production change
reverted, and observed failing.

### What is worth checking outside a review

- The margin rule changes how multi-line errors and `gwm doctor` render. Worth
  one look at a real terminal, which is the kind of information no further
  review pass produces.
- `labels::validate_label_name` guards `is_ascii_control` only, so the C1
  range passes it, and milestone titles are validated by nothing. Both are now
  neutralised at display; the validators themselves are untouched and
  pre-existing.

PR #497, opened 2026-08-03
🐛 fix(worktree): roll back a branch `add` created when the worktree fails

### LINKED ISSUE #487: worktree::add creates the branch before the directory, so any late failure leaves it orphaned
## Description

`worktree::add` (`src/worktree.rs:324`) creates the branch before it creates the worktree directory:

```rust
Err(_) => (repo.branch(branch_name, &head_commit, false)?, true),
...
repo.worktree(name, target_path, Some(&opts))?;
```

Anything that makes `repo.worktree` fail therefore leaves a branch behind that the user never asked for and did not create. It points at the old HEAD, it shows up in `git branch`, and `gwm doctor`'s orphan check is what eventually notices. Recovering means knowing to run `git branch -D` on a name the failed command printed in an error, not in a hint.

## Why now

This came out of #475 and was deliberately left out of its PR (#486), because it is the wider fix rather than that one. #475 closed one input set by refusing the name up front. The ordering problem is orthogonal: it fires on *every* late failure, not on a name.

Two of these have been closed one input set at a time already:

- #474 capped the name at 255 bytes, because `a×130/b×130` is a legal ref and an illegal directory name, and the branch was being created before the directory failed.
- #475 refused the Windows character and reserved-device set for the same reason.

Both are real fixes and both leave the shape intact. A third input set will find it again, and the classes that remain are not ones a validator can enumerate: a full disk, a permission denial on the base directory, a `target_path` on a filesystem that refuses a name the local one accepts, an `.git/worktrees/<name>` entry left by a crashed earlier run.

## Proposed solution

`add` already tracks whether it created the branch:

```rust
let (branch, created_branch) = match repo.find_branch(...) { ... };
```

`created_branch` is currently used only to stamp `branch.<name>.gwm-created-at`. Roll the branch back on the `repo.worktree` failure path when, and only when, it is `true`. A branch that was reused (`--reuse-branch`, `created_branch == false`) must be left alone: it existed before the command ran, and deleting it would destroy work.

The error the user sees should stay the underlying one. The rollback is not the story, it just stops a failed create from leaving state behind.

## Alternatives considered

**Create the worktree first, then the branch.** Not available: `WorktreeAddOptions::reference` needs the reference to exist before `repo.worktree` is called, which is what makes the current order the natural one rather than an oversight.

**Leave it and let `gwm doctor` report it.** That is today's behaviour. It works, but it makes a failed `gwm create` a two-step recovery, and the orphan check is opt-in rather than something a user runs after every failure.

**Keep enumerating input sets in the validator.** #474 and #475 are both worth having on their own merits (a name refused up front beats a name refused halfway through, whatever the cleanup). But they cannot cover a full disk.

## Acceptance

- [ ] a `repo.worktree` failure after `add` created the branch leaves no branch behind
- [ ] a `repo.worktree` failure on a *reused* branch (`--reuse-branch`) leaves it untouched
- [ ] the error surfaced to the user is still the underlying failure, not a rollback message
- [ ] covered by an integration test in `tests/worktree_integration.rs` that forces `repo.worktree` to fail after the branch is created, rather than by an input that a validator now rejects up front

kbrdn1:
## Description

`worktree::add` creates the branch before it creates the worktree, because
`WorktreeAddOptions::reference` takes a reference that already exists. Every
failure past that point therefore left a branch nobody asked for, pointing at
the old HEAD, surfaced only by `gwm doctor`'s orphan check or by a
`git branch -D` on a name the error printed in passing.

The branch is now deleted on the failure path, and only when this call created
it, still points where this call put it, and no checkout is standing on it.

Closes #487

## Type of change

- [x] 🐛 Fix (bug fix)
- [ ] 📝 Docs
- [ ] ✅ Tests
- [ ] ⚡ Perf
- [ ] 🔧 Chore / CI / build

## Changes

- `worktree::add` rolls the branch back when `repo.worktree` fails, keyed on
  the `created_branch` flag the function already tracks for its created-at
  stamp. A reused branch predates the command and is left alone.
- The rollback deletes the reference and removes its own `gwm-created-at`
  stamp, rather than deleting the branch, which would take the whole
  `branch.<name>` config section with it.
- `branch_is_checked_out_anywhere` holds the rollback back when any checkout
  has the branch as its HEAD, which is the protection `git_branch_delete`
  carries and `git_reference_delete` does not.
- Eight integration tests in `tests/worktree_integration.rs`, red first.
- CHANGELOG entry under `[Unreleased] > Fixed`.

## Tests

- [x] `cargo test` passes locally (under a CI-like minimal PATH:
      `PATH="$(dirname "$(command -v cargo)"):/usr/bin:/bin" cargo test`)

## Screenshots / TUI captures

Not applicable, no TUI surface touched.

## Checklist

- [x] README updated if CLI surface changed (no CLI surface change)
- [x] `examples/gwm.toml.example` updated if config schema changed (no schema
      change)
- [x] No `unwrap()` on user-facing paths (return `GwmError` instead)
- [x] No `println!` in TUI render code (status bar only)

## Linked issues / docs

- Issue: #487
- Related PR: #486 (#475, where this was deliberately left out as the wider
  fix), #474

## Notes for reviewers

**Three conditions gate the delete, and each one is a test.** The rollback
fires only when this call created the branch, the tip still matches the commit
it was created at, and no checkout has it as HEAD. The three came from three
different failure modes, all of them worse than the orphan branch the PR is
about.

**The predicate is `created_branch`, not `!reuse_branch`.** `gwm undo`
(`src/cli.rs`) and `gwm review` (`src/review.rs`) both call `add` with
`reuse_branch: true` against a branch that is usually absent, so `add` creates
it. A rollback keyed on `!reuse_branch` would leave exactly those two paths
orphaning branches, and it would pass a test pair built from "reuse off,
created" plus "reuse on, pre-existing".
`add_rolls_back_a_branch_it_created_even_when_reuse_was_asked_for` is the third
case that separates them.

**The ref goes, not the branch.** `git_branch_delete` drops the whole
`branch.<name>` config section, and that section outlives its ref easily:
`git update-ref -d` leaves it behind, and an upstream can be configured before
the branch exists. So the rollback deletes the reference and removes only the
`gwm-created-at` key `add` itself wrote. Verified: libgit2 does remove the
section on `git_branch_delete` (branch.c, `git_config_rename_section` right
before `git_reference_delete`), which is why this had to change.

**That trade cost a protection, and the third condition buys it back.**
`git_branch_delete` refuses a branch that is "the current HEAD of a linked
repository"; `git_reference_delete` does not, and git2 binds neither the check
nor `git_branch_is_checked_out`. Measured on a live worktree rather than
assumed: the ref goes, `find_worktree` still resolves the entry, and
`prunable_worktrees` lists nothing, so the residue is a worktree bound to a ref
that no longer exists and nothing reports it. `branch_is_checked_out_anywhere`
reads the main `HEAD` and every `worktrees/*/HEAD` under `commondir`, from disk
rather than through `repo.worktrees()`, because a half-written entry has no
`gitdir` file and libgit2 will not list it while its `HEAD` still names a
branch. A read that fails counts as in use; only a missing file counts as free.

**Where the rollback stops.** libgit2 writes `.git/worktrees/<name>/HEAD` near
the end of `git_worktree_add`, just before the checkout, so a checkout failure
lands with the worktree already bound and the branch is deliberately kept.
Behaviour for that window is exactly what it was before this PR. Everything
earlier is rolled back, which is where the disk-full and permission classes
land.

**Why the fixtures are what they are.** None of them uses an input a validator
could refuse up front, since that is the shape #474 and #475 already closed
twice and the point of this issue is the ordering. They are a
`.git/worktrees/<name>` entry left by a crashed earlier run (libgit2 mkdirs it
with `EXCL`), a permission denial on the target's parent, and hand-written
admin `HEAD` files for the bound cases, since a checkout failure is not
reachable from an integration test. The two permission-based tests are
`#[cfg(unix)]` and skip explicitly, with a message, when the process can write
to a `0555` directory or read a `0000` one, so running as root does not turn
them into silent passes.

**Prove-red.** Every one of the eight tests was run against the code without
its fix and failed on the assertion it is named for, not on a precondition. The
three original rollback tests failed on "branch left behind" specifically,
which rules out a fixture short-circuiting before the branch is created.
`add_leaves_a_reused_branch_alone_when_the_worktree_fails` was green from the
start, as it should be.

**One thing is untestable from the public surface**: the tip comparison needs
the ref to move *inside* `repo.worktree`, and there is no seam for that. It is
argued here under the narrow exception in CLAUDE.md rather than left silent.

**Not in scope.** libgit2 leaves its own `.git/worktrees/<name>` admin entry
behind on the failure paths that get that far. `gwm prune` and
`prunable_worktrees` already handle orphaned admin entries, and cleaning it
here would mean deciding what to do with a half-created worktree directory too.
Worth a separate issue if it bites.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -173,6 +173,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- A `gwm create` that fails while making the worktree no longer leaves the
+  branch behind. `worktree::add` has to create the branch before it calls
+  `repo.worktree`, because libgit2's `WorktreeAddOptions::reference` takes a
+  reference that already exists, so any failure past that point left a branch
+  nobody asked for, pointing at the old HEAD, surfaced only by `gwm doctor`'s
+  orphan check or by a `git branch -D` on a name the error printed in passing.
+
+  It is now deleted on the failure path, and only when that same call created
+  it: a branch attached with `--reuse-branch` predates the command, and
+  deleting it would destroy work. The predicate is "this call created it"
+  rather than "reuse was off", because `gwm undo` and `gwm review` both pass
+  the reuse flag against a branch that is usually absent, and then create it.
+  libgit2 drops the `branch.<name>` config section along with the ref, so the
+  creation stamp goes with it instead of ageing the next branch to take that
+  name.
+
+  The error stays the underlying failure rather than a cleanup message. One
+  residue is out of reach and remains `gwm doctor`'s job: libgit2 refuses to
+  delete a branch that is already the HEAD of a linked repository, which is
+  the state it leaves when the failure came from the checkout itself rather
+  than from either mkdir.
+
+  This bounds the class that #474 (the 255-byte cap) and #475 (the Windows
+  character and device set) each closed one input set of. Refusing a name up
+  front is worth having on its own, but a full disk is not an input set.
+
 - Config values no longer reach the terminal with their control bytes intact.
   A `.gwm.toml` comes from a repo nobody has vetted, and the commands that read
   it back skip the trust gate on purpose, because inspecting an unfamiliar repo
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -185,9 +185,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   deleting it would destroy work. The predicate is "this call created it"
   rather than "reuse was off", because `gwm undo` and `gwm review` both pass
   the reuse flag against a branch that is usually absent, and then create it.
-  libgit2 drops the `branch.<name>` config section along with the ref, so the
-  creation stamp goes with it instead of ageing the next branch to take that
-  name.
+  The creation stamp `add` wrote goes too, so it cannot age the next branch to
+  take that name, but the rest of the `branch.<name>` section stays put: that
+  section outlives its ref easily, and deleting the branch rather than the ref
+  would have taken an upstream setting the command never wrote.
 
   The error stays the underlying failure rather than a cleanup message. One
   residue is out of reach and remains `gwm doctor`'s job: libgit2 refuses to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -191,10 +191,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   would have taken an upstream setting the command never wrote.
 
   The error stays the underlying failure rather than a cleanup message. One
-  residue is out of reach and remains `gwm doctor`'s job: libgit2 refuses to
-  delete a branch that is already the HEAD of a linked repository, which is
-  the state it leaves when the failure came from the checkout itself rather
-  than from either mkdir.
+  window is left alone on purpose: once libgit2 has bound the worktree to the
+  branch, which it does just before the checkout, the branch stays. Deleting
+  it there leaves a worktree pointing at a ref that no longer exists, and
+  unlike an orphan branch that residue is reported by nothing, so a failure
+  that late behaves as it did before.
 
   This bounds the class that #474 (the 255-byte cap) and #475 (the Windows
   character and device set) each closed one input set of. Refusing a name up
--- a/src/worktree.rs
+++ b/src/worktree.rs
@@ -385,9 +385,17 @@ pub fn add(
     // from either mkdir; `gwm doctor` still catches that residue. The
     // caller gets the underlying error in every case, since the rollback
     // is not the story.
+    //
+    // The tip is re-checked because "what this call created" is a claim
+    // about an OID, not about a name: another process is free to move the
+    // ref while `repo.worktree` runs, and a branch that no longer points
+    // where this call put it is somebody else's now. Leaving an orphan is
+    // the smaller harm of the two.
     if created_branch {
       if let Ok(mut b) = repo.find_branch(branch_name, git2::BranchType::Local) {
-        let _ = b.delete();
+        if b.get().target() == Some(head_commit.id()) {
+          let _ = b.delete();
+        }
       }
     }
     return Err(e.into());
--- a/src/worktree.rs
+++ b/src/worktree.rs
@@ -379,12 +379,13 @@ pub fn add(
     // asked for (#487). Roll back only what *this* call created: a reused
     // branch predates the command, and deleting it would destroy work.
     //
-    // Best effort, and deliberately so. Once libgit2 has bound the
-    // worktree to the branch, which it does just before the checkout, the
-    // branch stays: deleting it there leaves a worktree pointing at
-    // nothing, and that residue is reported by nothing at all, unlike an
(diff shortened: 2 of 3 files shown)


kbrdn1:
## Codex review loop: stopped on stagnation at pass 6/8, not on a clean check

Last `P0P1P2_COUNT` seen in the real output is **1**, and that finding is
refuted below rather than fixed. Five findings were taken and are in the
branch; the sixth asks for a guarantee git itself does not provide.

| pass | findings | P0 | P1 | P2 | what it pointed at |
|:--|:--|:--|:--|:--|:--|
| 1 | 1 | 0 | 1 | 0 | the rollback looked the branch up by name, so a concurrent `git branch -f` in the window would have had its work deleted |
| 2 | 1 | 0 | 0 | 1 | `Branch::delete` drops the whole `branch.<name>` config section, including settings the command never wrote |
| 3 | 1 | 0 | 0 | 1 | pass 2's code: `Reference::delete` carries none of `git_branch_delete`'s linked-HEAD protection |
| 4 | 1 | 0 | 1 | 0 | pass 3's code: the guard was keyed on the admin entry name, not on the branch |
| 5 | 1 | 0 | 1 | 0 | pass 4's code: a failed read counted as "not checked out", so the guard failed open |
| 6 | 1 | 0 | 1 | 0 | pass 5's code: the TOCTOU window between the check and the delete |

**Where it turned.** Pass 3. Passes 1 and 2 pointed at the rollback as
originally written; every pass from 3 on pointed at a line written to answer
the pass before it. Passes 3, 4 and 5 were still worth taking, because each one
named a way the guard was wrong rather than a scenario it did not cover, and
pass 3 in particular caught a protection I had silently dropped while fixing
pass 2. Pass 6 is where that stops being true.

**Severity trajectory.** Worst finding of pass 1: the rollback could delete a
branch another process had just moved, which is data loss. Worst finding of
pass 6: a window of a few microseconds between reading the HEAD files and
deleting the ref.

**The remaining finding, refuted with primary evidence.** It asks to
"serialise the HEAD check and the delete, or give up the rollback when that
guarantee cannot be obtained". Git provides no lock over another worktree's
`HEAD`, and libgit2 does the same check-then-act in both of the functions this
code stands in for:

- `git_branch_delete` (`branch.c:212`): `git_branch_is_checked_out(branch)`,
  then `git_config_rename_section`, then `git_reference_delete(branch)` at
  line 226. No lock between them.
- `git_worktree_add` (`worktree.c:348`): `git_branch_is_checked_out(ref)`,
  then it goes on to bind the worktree.

So the window this finding describes is the window `git branch -d` has had all
along, on a branch this call created seconds earlier, in the interval between
two adjacent statements. Taking the second half of its own suggestion, giving
up the rollback, means reverting the PR. Not actionable.

**What is worth checking outside the review.** None of this has run against a
real failure: the tests reach `repo.worktree`'s error path through a wedged
admin entry and a read-only directory, which is the honest way to reach it, but
a genuinely full disk sequences the failure differently and would land in the
checkout stage, where the branch is now deliberately kept. That window is
unchanged by this PR and is the thing to watch if it ever shows up in an issue.

Issue #491, opened 2026-08-03
[Task]: declared MSRV is 1.86 but the locked graph requires 1.88 (ratatui 0.30 / time) + no CI guard


kbrdn1:
### Scope

`Cargo.toml` declares `rust-version = "1.86"`, but the locked dependency graph
requires **1.88**. The declared MSRV has been wrong since ratatui 0.30 landed.

```
$ cargo metadata --format-version 1 --locked \
  | jq -r '.packages[] | select(.rust_version != null) | "\(.rust_version)\t\(.name) \(.version)"' \
  | sort -Vr | head
1.88.0	time-core 0.1.8
1.88.0	time 0.3.47
1.88.0	ratatui-widgets 0.3.2
1.88.0	ratatui-termwiz 0.1.2
1.88.0	ratatui-termina 0.1.0
1.88.0	ratatui-crossterm 0.1.2
1.88.0	ratatui-core 0.1.2
1.88.0	ratatui 0.30.2
1.88.0	darling_macro 0.23.0
1.88.0	darling_core 0.23.0
...
1.86	gwm-cli 1.5.0        <- Cargo.toml
```

With `Cargo.lock` committed, cargo refuses at resolve time on a 1.86 toolchain
(`package 'ratatui-core v0.1.2' cannot be built because it requires rustc 1.88.0
or newer`). This is cargo's own `rust-version` gate, not a compilation failure,
so it fires before a single crate is built.

Two parts, and they only work together:

**1. Bump the declared MSRV to 1.88**, everywhere it is claimed:

- [ ] `Cargo.toml:21` `rust-version = "1.88"` + rewrite the explanatory comment
      above it (it currently attributes the floor to `tui-term 0.3.4`, which
      declares 1.86; the real driver is now the ratatui 0.30 stack)
- [ ] `README.md:14` rust badge
- [ ] `CONTRIBUTING.md:115`
- [ ] `ROADMAP.md:50`
- [ ] `docs/1.getting-started/1.install.md:18` + `docs/fr/…`
- [ ] `docs/6.development/3.stability.md:102-117` + `docs/fr/…`
- [ ] `docs/7.roadmap.md:12` + `docs/fr/…`

**2. Add an MSRV job to CI** so it cannot drift silently again:

```yaml
msrv:
  runs-on: ubuntu-latest
  steps:
    - uses: actions/checkout@v4
    - uses: dtolnay/rust-toolchain@1.88
    - run: cargo check --all-targets --locked
```

[...]

PR #499, opened 2026-08-03
⬆️ chore(deps): refresh the transitive lockfile

kbrdn1:
## Description

`cargo update` on the whole graph. `.github/dependabot.yml` sets no
`allow: dependency-type: all`, so dependabot only PRs the direct dependencies
and the transitive graph drifts between manual refreshes. Lockfile only.

No linked issue: same as the dependabot PRs this stands in for.

## Type of change

- [ ] 🐛 Fix (bug fix)
- [ ] 📝 Docs
- [ ] ✅ Tests
- [ ] ⚡ Perf
- [x] 🔧 Chore / CI / build

## Changes

- 60-odd crates moved, 12 dropped that nothing needs any more (the
  `wit-bindgen` / `wasm-encoder` / `wit-component` chain, `prettyplease`,
  `leb128fmt`, `unicode-xid`).
- `libgit2-sys` 0.18.4+1.9.3 to 0.18.7+1.9.6, so the **vendored libgit2** takes
  three patch releases.
- `time` 0.3.47 to 0.3.55, one of the crates that pinned the MSRV floor
  discussion in #491.
- `shlex` crosses a major, 1.3.0 to 2.0.1, as a **build-only** dependency of
  `cc`, which the `libgit2-sys` / `libsqlite3-sys` / `libz-sys` build scripts
  pull. Nothing links it into the binary.

## Tests

- [x] `cargo test` passes locally (under a CI-like minimal PATH:
      `PATH="$(dirname "$(command -v cargo)"):/usr/bin:/bin" cargo test`)
- [x] `cargo fmt --check` passes (no source change)
- [ ] New tests added under `tests/` (lockfile only, no behaviour to pin)

## Screenshots / TUI captures

Not applicable.

## Checklist

- [ ] Branch follows `<type>/#<issue>-<description>` (free-form name, no issue,
      matching how dependency bumps land here)
- [ ] CHANGELOG.md updated under `## [Unreleased]` (dependency bumps do not get
      an entry here; the merged dependabot PRs touch `Cargo.lock` alone)
- [x] README updated if CLI surface changed (no surface change)
- [x] `examples/gwm.toml.example` updated if config schema changed (no schema
      change)
- [x] No `unwrap()` on user-facing paths (no source change)
- [x] No `println!` in TUI render code (no source change)

## Linked issues / docs

- Issue: none
- Related PR: #498 (the MSRV floor this had to be checked against)

## Notes for reviewers

**The MSRV was checked, not assumed.** A lockfile refresh is exactly how a
floor moves without anyone deciding to, which is the #491 bug in its natural
habitat. Ran the #498 CI job locally against the new lockfile:
`cargo +1.95 check --all-targets --locked` and the same with
`--no-default-features`, both green. `--locked` is the load-bearing part: the
resolve-time gate fires for crates that declare a floor, and the compile that
follows is what catches one declaring none.

`cargo audit --deny warnings` also ran on the new lockfile: 306 crates
scanned, clean.

**Timing.** Deliberately after #498 merged rather than before. The floor is
declared at 1.95 there; refreshing the graph first would have moved the target
while it was being measured.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+checksum = "c982642fa9e8606056828ee9a8505737230110bb1099153c79efe865c59d12ba"
 dependencies = [
  "memchr",
 ]
@@ -108,9 +108,9 @@ dependencies = [
 
 [[package]]
 name = "arrayvec"
-version = "0.7.6"
+version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
+checksum = "d3fb67a6e08acf24fdeccbac2cb6ac4305825bd1f117462e0e6f2f193345ad56"
 
 [[package]]
 name = "assert_cmd"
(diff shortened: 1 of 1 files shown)


kbrdn1:
CI went red once on `test (ubuntu-latest)` and it is not this PR. `exec_in_dir_runs_a_relative_script_from_the_worktree` failed with `ETXTBSY` while macOS and Windows were green on the same commit, and a re-run of that job alone, with nothing changed, went green. Filed as #500 with the analysis: writing an executable and running it from a multi-threaded harness races against the fork-to-exec window of any other test that spawns a process, so the kernel calls the text file busy. Nothing to do with the lockfile.

All 13 checks green now, including the three `msrv (declared floor, ...)` jobs that #498 added, which is the one thing a lockfile refresh can genuinely break.

Issue #503, opened 2026-08-03
[Bug]: five config tests read $HOME concurrently with a set_var that takes no lock


kbrdn1:
## Description

`tests/config_tests.rs` has an `env_lock` whose doc says it guards tests that
expand a `{home}` pattern. That is the wrong boundary. `expand_placeholders`
reads `$HOME` unconditionally, before it looks at any token:

```rust
// src/config.rs:2071
pub fn expand_placeholders(...) -> Result<String> {
  let home = dirs::home_dir()
    .ok_or_else(|| GwmError::Config("cannot resolve $HOME".into()))?
```

So every test in that binary calling `expand_placeholders` is a concurrent
reader of `$HOME`, whether or not its template mentions `{home}`. Meanwhile
`a_token_produced_by_expanding_home_stays_literal` calls
`std::env::set_var("HOME", ...)`, and these do not take the lock:

- `placeholders_expand_repo_path_and_parent`
- `placeholders_repo_path_tokens_left_literal_without_path`
- `a_token_produced_by_expanding_the_repo_name_stays_literal`
- `a_token_with_no_value_survives_the_single_pass_verbatim`
- `a_repo_path_without_a_parent_resolves_only_the_path_token`

`set_var` is `unsafe` precisely because it races a concurrent `getenv`. The
failure mode is therefore a rare flake or a crash rather than a wrong
assertion, and a mutex only serialises the participants that take it.

Raised by CodeRabbit on the v1.6.0 release PR (#501); the `home_dir` call was
verified in the source rather than taken on trust. Same family as #500: a
latent race in the harness, not in the product.

## Steps to reproduce

Not reliably reproducible on demand, which is the point. `cargo test --test
config_tests` runs these on several threads in one process, and one of them
mutates `HOME` while the others read it.

## Expected behaviour

Every test that can observe `$HOME` is serialised against the one that
rewrites it.

## Actual behaviour

Five of them are not.

## Severity

🟡 Minor (annoying but can be worked around)

## Proposed fix

Take the guard in the five tests above, and widen the helper's doc comment from
"expands a `{home}` pattern" to "calls `expand_placeholders`", so the next test
to be added inherits the right rule rather than the narrow one. Worth checking
the other test binaries for the same shape while in there.

kbrdn1:
Fixed by #505, merged into `dev` as `86510d9`. Closing by hand: a PR merged into `dev` does not trigger the `Closes` keyword, which only fires on the default branch.

PR #509, opened 2026-08-04
✅ test(config): check the env-guard rule by construction

### LINKED ISSUE #507: [Bug]: the $HOME guard in config_tests stops at expand_placeholders, and the other test binaries have the same shape
## Description

#503 took the `env_lock` guard in the five tests that call
`expand_placeholders` without it, and rewrote the doc to say the rule is "can
this test observe `$HOME`?" rather than "does this pattern contain `{home}`?".
That is the right rule and the change does not yet satisfy it.

A layered load reads `$HOME` too:

```
Config::load_layered → global_config_path() → dirs::home_dir()   // src/config.rs
```

`tests/config_tests.rs` does not set `GWM_NO_GLOBAL_CONFIG` the way
`terminal_escape_tests.rs` does through its `gwm_in` helper, so every
`load_layered` test in that binary is also a concurrent reader of `$HOME`,
racing the `set_var` in `a_token_produced_by_expanding_home_stays_literal`.

It was left out of #503 on purpose: taking the lock in all of them would
serialise most of a 127-test binary to fix a race whose window is one test.

## Steps to reproduce

Not reproducible on demand, same as #503. `set_var` is `unsafe` because it
races a concurrent `getenv`; the failure mode is a rare flake or a crash.

## Expected behaviour

Either every test that can observe `$HOME` is serialised against the one that
rewrites it, or the tests that can observe it are made not to.

## Actual behaviour

Only the `expand_placeholders` callers take the guard.

## Severity

🟡 Minor (annoying but can be worked around)

## Two directions, and the second is probably better

1. **Widen the lock** to the `load_layered` tests. Correct, and it serialises
   most of the binary.
2. **Remove the exposure instead**: pin the global layer off for the tests
   that do not care about it (`GWM_NO_GLOBAL_CONFIG`, the mechanism
   `terminal_escape_tests` already uses), so they stop reading `$HOME` at all
   and the lock stays small. A test that does not touch the variable cannot
   race it.

## The other binaries

#503 noted these are worth checking. The check was run and came back
inconclusive: `trust_tests` (`GWM_TRUST_LEDGER`), `config_global_tests`
(`XDG_CONFIG_HOME`, `GWM_NO_GLOBAL_CONFIG`) and `history_tests`
(`GWM_HISTORY_FILE`) each mutate a variable, but distinguishing a test that
really reads it from one that passes an explicit path needs reading the
helpers rather than grepping for names. That reading is the other half of this
issue.

Note each test binary is its own process, so a lock in one says nothing about
another, and `naming_tests` calls `expand_placeholders` without any guard yet
mutates no environment variable at all, which is fine: readers without a
writer do not race.

kbrdn1:
## Description

#507 as filed describes a race that does not exist. The premise was verified
before any code was written and it does not hold, so what this PR delivers is
the part that survives: the rule has now been stated wrongly twice in a doc
comment, so it stops being prose and becomes a test.

Closes #507

## Type of change

- [x] ✅ Tests
- [x] 📝 Docs

## Changes

- `tests/env_guard_invariant_tests.rs`: walks the four test binaries that
  rewrite an environment variable and fails on any test that calls an ambient
  reader of that variable without holding the binary's lock. A companion test
  exercises the matcher in both directions so the guard cannot pass by
  matching nothing.
- `tests/config_tests.rs`: `env_lock`'s doc no longer claims a layered load
  reads `$HOME`.
- `CHANGELOG.md` under `## [Unreleased]`.

## Tests

- [x] `cargo test` passes locally (91 binaries green)

Run under a CI-like stripped environment too, since this is entirely about
environment-dependent tests:

```
PATH="$(dirname "$(command -v cargo)"):/usr/bin:/bin" cargo test   # exit 0
```

## Checklist

## Linked issues / docs

- Issue: #507, with the correction posted as a comment there.
- Follows #505, which is where the wrong doc comment was introduced.

## Notes for reviewers

**Why the issue was wrong, since the PR is shaped by it.**
`Config::load_layered(repo_root, global_path)` takes the global config path as
a parameter and resolves nothing; `Config::load_for_repo` is the one that calls
`global_config_path()`. No test in `config_tests.rs` calls it: of six
occurrences of the name, five are comments and the sixth is a test's own name.

**The audit is the test.** An *ambient reader* is a function that resolves the
variable itself, and telling one from a pure function that takes paths as
parameters is the entire exercise. Both naive searches failed, in opposite
directions:

- substring on `load_for_repo` counts five comments as calls;
- substring on `global_config_path` reports every `resolve_global_config_path`
  test as an offender.

The matcher therefore skips comment lines and requires a non-identifier
character before the name and a `(` after it, and `the_guard_can_actually_fire`
pins both.

**Red proven.** Deleting the `env_lock()` line from one real test makes the
invariant fail naming that test and its file. Restored, 91 binaries are green.

**What it does not do.** It does not detect a reader reached indirectly, only a
direct call to a listed name. That is a deliberate floor: the list of ambient
readers is short, verified against `src/`, and carries a comment saying what
qualifies, so extending it is a one-line change when a new one appears.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- The rule that a test which can observe a rewritten environment variable must
+  hold its binary's lock is now checked by construction rather than stated in a
+  doc comment. That comment is what #503 was about: it named a boundary
+  narrower than the real one, and five tests were written against the sentence.
+  Its replacement then stated a second wrong boundary, that a layered load
+  reads `$HOME`. It does not: `Config::load_layered` and `resolved_rows` take
+  the global config path as a parameter, and `Config::load_for_repo`, which
+  does resolve it, is called by no test in that binary. A new test walks the
+  four test binaries that rewrite a variable and fails on any test calling an
+  ambient reader of it without the lock, ignoring mentions in comments and
+  identifiers that merely end with the name. Both mistakes came from a search
+  that could not tell those apart.
+  ([#507](https://github.com/kbrdn1/gwm-cli/issues/507))
+
 - A branch name no longer reaches the TUI table with its bidi controls intact.
   The sinks the CLI goes through are not on the TUI's path, and what protected
   the TUI so far was incidental: measured on ratatui 0.30, every render path
--- a/tests/config_tests.rs
+++ b/tests/config_tests.rs
@@ -16,9 +16,18 @@ use tempfile::TempDir;
 /// adding a test can answer, rather than "does this string contain `{home}`?",
 /// which looks answerable and is not.
 ///
-/// A layered load reads it too (`Config::load_layered` → `global_config_path`
-/// → `dirs::home_dir`). Those tests do not take the lock yet; widening it
-/// there would serialise most of the binary, so it is tracked separately.
+/// The other half of that rule is which functions can see `$HOME`, and the
+/// note here used to answer it wrongly (#507): it claimed a layered load reads
+/// it through `Config::load_layered`. It does not. `load_layered` and
+/// `resolved_rows` take the global config path as a **parameter**, so they
+/// resolve nothing; `Config::load_for_repo` is the one that calls
+/// `global_config_path()`, and no test in this binary calls it. So
+/// `expand_placeholders` is the only ambient reader reachable from here, and
+/// #503 covered it completely.
+///
+/// The rule is no longer only stated: `tests/env_guard_invariant_tests.rs`
+/// checks it by construction, for this binary and the three others that
+/// rewrite an environment variable.
 fn env_lock() -> &'static std::sync::Mutex<()> {
   static LOCK: std::sync::OnceLock<std::sync::Mutex<()>> = std::sync::OnceLock::new();
   LOCK.get_or_init(|| std::sync::Mutex::new(()))
--- a/tests/env_guard_invariant_tests.rs
+++ b/tests/env_guard_invariant_tests.rs
@@ -1,0 +1,148 @@
+//! One invariant, checked by construction rather than stated in a doc
+//! comment (issue #507).
+//!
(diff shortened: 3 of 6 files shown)


kbrdn1:
## Review loop stopped on stagnation, not on clean

Five passes of the local Codex review, every one of them on this file. Recording
where it landed, because the last two findings are declined rather than fixed.

| pass | findings | P1 | P2 | what they pointed at |
|---|---|---|---|---|
| 1 | 3 | 0 | 3 | the first version, three hand-maintained lists |
| 2 | 4 | 0 | 4 | the code written to fix pass 1 |
| 3 | 2 | 0 | 2 | the code written to fix pass 2 |
| 4 | 2 | 1 | 1 | the code written to fix pass 3 |
| 5 | 3 | 0 | 3 | the code written to fix pass 4 |

**Eleven were real and are fixed**, including one that mattered: in pass 3 the
unique-name filter was dropping `expand_placeholders` (defined in both
`config.rs` and `lifecycle.rs`), so the guard stayed green with the lock deleted
from the exact test #503 was about. That one is now pinned by its own test
against the real `src/`.

**Two are declined**, both documented at the code with the measurement behind
them:

- *Readers that take the variable name as a parameter* (`env_truthy(key)`).
  Seeding on non-literal `env::var` calls marks it a reader of everything;
  through the transitive walk that demands a lock from about a third of
  `forge_tests` and `tui_app_tests`, including binaries whose rewritten-variable
  set is empty. Measured, not assumed.
- *Guard scope* (a lock taken in an inner block, or dropped before the read).
  Following scope means tracking blocks. No audited binary takes its lock in an
  inner block, and the only `drop(` in one of them drops a repo handle.

Pass 5's remaining two are the same shape: block comments containing Rust that
looks like a call, and direct `std::env::var` reads written inline in a test.
Neither exists in the tree.

**Why stop here.** The findings have pointed at the previous pass's code for
five passes running, the severity has not moved off P2 except once, and the
last ones describe forms nobody has written. What the guard claims is now
written in its module doc, including what it does not check. The next useful
information about this invariant comes from a test that trips it, not from
another pass.

PR #510, opened 2026-08-04
👷 ci: ring the docs site bell when main moves

kbrdn1:
## Description

The `gwm-cli` half of #423: this repo owns the Markdown sources and tells the
docs site when they move. The conversion, the Starlight build and the deploy
all live in `kbrdn1/kbrdn-docs`.

**This does nothing on its own.** The dispatch it sends is only acted on once
`sync-gwm.yml` is on `main` of `kbrdn-docs`. Merging this early is deliberate:
the two halves are independent, and an unanswered dispatch is a no-op.

refs #423

## Type of change

- [x] 🔧 Chore / CI / build

## Changes

- `.github/workflows/docs-sync.yml`: on a push to `main` touching `docs/` or
  `changelogs/`, sends a `repository_dispatch` to `kbrdn1/kbrdn-docs`. No
  checkout, no permissions, one API call.
- `tests/release_workflow_tests.rs`: the credential sweep discovers the
  workflow files instead of naming two of them, and two new tests pin the
  trigger and the paths filter.
- `CONTRIBUTING.md`: a `Documentation site` section under Releases, with the
  one-time token bootstrap.

## Tests

- [x] `cargo test` passes locally (91 binaries green)

## Checklist

- [ ] CHANGELOG.md updated under `## [Unreleased]`

No CHANGELOG entry: nothing here changes the CLI, its output or its config.
Say the word if you would rather see it under Added anyway.

## Linked issues / docs

- Issue: #423 (this closes the `gwm-cli` half only, so `refs` rather than
  `Closes`)

## Notes for reviewers

**Why `main` and not a tag.** GitHub's `v*.*.*` glob matches `v0.8.0-rc.4`
too. `release.yml` already has to guard against that by hand, and a docs
trigger with the same glob would put every release candidate's pages online as
if they were stable. `main` is only ever reached through a `dev` → `main` PR,
so it *is* the delivered state.

**The two new tests were watched failing**, each on its own assertion: pointing
`branches` at `dev` fails the trigger test and leaves the paths test green;
deleting `changelogs/**` does the reverse.

**The sweep now has a floor that bites.** Deleting `docs-sync.yml` drops the
workflow count to 2 and fails the test, which is the difference between a sweep
and a loop that happens to find nothing. Without this change the new workflow
would have been outside a credential invariant meant to be repo-wide, and so
would the one after it.

**One credential note.** `DOCS_SITE_TOKEN` is a write credential for a private
repository, held in a public repository's secrets. Same trust level as
`HOMEBREW_TAP_TOKEN` and `SCOOP_BUCKET_TOKEN`; it is scoped to
`kbrdn1/kbrdn-docs` with Contents: write and nothing else. It is already
configured, so the `continue-on-error: true` on the notify job should come off
once the first dispatch is seen green.

### DIFF
--- a/.github/workflows/docs-sync.yml
+++ b/.github/workflows/docs-sync.yml
@@ -26,6 +26,12 @@ on:
     paths:
       - 'docs/**'
       - 'changelogs/**'
+      # Not documentation, but the site reads it: the sync script pulls the
+      # `[package]` version out of it for the release badge in the header.
+      # A release usually moves `changelogs/` in the same push, so this only
+      # matters when a version bump travels alone, which is exactly the case
+      # nobody would think to check.
+      - 'Cargo.toml'
   workflow_dispatch:
 
 # The dispatch authenticates with a PAT scoped to the docs repository; the
--- a/.github/workflows/docs-sync.yml
+++ b/.github/workflows/docs-sync.yml
@@ -1,0 +1,65 @@
+name: docs-sync
+
+# The published documentation follows `main`, never `dev`.
+#
+# `main` is reached exclusively through a `dev` → `main` pull request, so what
+# lands here is what was delivered — which is exactly what the site should
+# serve. Triggering on `dev` would publish pages describing behaviour nobody
+# can install yet; triggering on tags would be worse still, since GitHub's
+# `v*.*.*` glob also matches `v0.8.0-rc.4` (the trap `release.yml` guards
+# against by hand) and every release candidate's docs would go live as stable.
+#
+# The split of work is deliberate: this repo owns the documentation sources
+# (`docs/`, `changelogs/`) and nothing else. The conversion script, the
+# Starlight site and the deploy live in `kbrdn1/kbrdn-docs`, so this workflow
+# only rings the bell. Doing the sync here would drag a Bun monorepo into the
+# CLI's CI for a build this repo cannot verify.
+#
+# Bootstrap: see CONTRIBUTING.md > Releases > Documentation site.
+
+on:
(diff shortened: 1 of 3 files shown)


kbrdn1:
Pushed three follow-ups: the Windows failure, plus both review findings, each verified against `kbrdn-docs` rather than taken on trust.

**`test (windows-latest)` was a real portability bug, not a flake.** `read_dir` joins with the platform separator, so the sweep compared `.github/workflows\\release.yml` against a literal `/` path, the exclusion missed, and it audited the one workflow that is *supposed* to carry a token. It now matches on the file name. Not on a suffix: `pre-release.yml` ends with `release.yml`, so the obvious fix would have dropped a workflow that must be audited, and the `swept` floor is what would have caught that.

**`Cargo.toml` was missing from the paths filter.** Confirmed in the source: `scripts/sync-gwm-docs.mjs:441` reads it for the `[package]` version behind the site header's release badge. A release moves `changelogs/` in the same push, so the gap only shows when a version bump travels alone and the symptom is a stale badge, never an error. The test claiming to pin every root the site reads now does. Watched failing with the entry removed.

**The receiver must be on the default branch**, and that is now a bootstrap step. Confirmed: `sync-gwm.yml` is present on `origin/dev` of `kbrdn-docs` and absent from `origin/main`. GitHub only runs a `repository_dispatch` receiver from the default branch, so until it lands there the API call returns 204, this workflow goes green, and nothing syncs, with no failure anywhere to notice. Which is the same thing this PR's description says about landing early, now written where someone will look for it.

Issue #98, opened 2026-05-22
[Bug]: gwm remove deletes worktree dir before prune — failed prune leaves phantom worktree


kbrdn1:
## Description

`worktree::remove` (verified at `src/worktree.rs:252-260`) deletes the working directory with `fs::remove_dir_all` **before** calling `git worktree prune` (or the libgit2 equivalent). If the `prune` step fails mid-way (libgit2 error, permissions, partial state), the result is a **phantom worktree**: the on-disk directory is gone, but `repo.worktrees()` still lists the worktree by name, so `gwm list` shows a ghost row, `gwm path` resolves to a non-existent directory, and `gwm bootstrap` against the ghost name fails with a confusing error.

The fix is the inverse order — prune first (administrative metadata), then `remove_dir_all` (filesystem) — so any failure leaves the worktree visibly present on disk and recoverable via `gwm bootstrap`.

## Steps to reproduce

Hard to repro deterministically (depends on a prune failure), but the source ordering at `worktree.rs:252-260` is the bug regardless. The smoke test:

```bash
gwm create feat 1 test
chmod -w "$(gwm path 1)/.git"        # force prune to fail (or simulate with libgit2 stub)
gwm remove feat-1
# expected: full rollback, worktree still present
# actual:   directory gone, repo.worktrees() still lists feat-1
gwm list                              # shows a phantom feat-1 row
```

## Expected behaviour

`worktree::remove` performs the prune (administrative removal) first. Only if that succeeds does it call `fs::remove_dir_all` on the worktree directory. A failure during prune leaves the worktree fully intact (filesystem + metadata) and surfaces a `GwmError` naming the failed step.

## Actual behaviour

`fs::remove_dir_all` runs first. A subsequent prune failure leaves the repo with a worktree entry whose backing directory has already been deleted. `repo.worktrees()` keeps returning the ghost name until `gwm prune` is run manually.

## Severity

🟠 Major (feature unusable, workaround painful — `gwm prune` recovers it but the user has to know)

## Surface

worktree (libgit2 ops)

## gwm version

0.6.0 (current `dev` HEAD)

## Rust version

n/a (source review finding — confirmed at `src/worktree.rs:252-260`)

## OS / shell

n/a (logic bug, OS-independent)

## Additional context

Tracked as **finding B / P1.6** in the multi-agent review (bug-hunt agent — unique finding, source-verified).

Fix:

[...]

PR #520, opened 2026-08-04
✨ feat(tui,cli): multi-row selection and bulk delete (#484)

### LINKED ISSUE #484: [Feature]: space as toggle the active row as to be "selected" e.g. for bulk cleanup
### Problem

I'd like to bulk cleanup worktrees with their branches.

### Proposed solution

For other tui tools I've made [e.g. gh-pr-lander](https://github.com/sassman/gh-pr-lander-rs/) I build bulk actions from the beginning. 
That makes me used to press a key (like space) to toggle selection and apply eventually an action to it. Here it would be select several worktrees and press `d` for deletion then.

### Alternatives considered

I see that space is bound already, but I'm suggesting an alternative key for the layout toggle, because: muscle memory. 

But of course open for suggestions of other keys to do the toggle.

### Surface

tui (new view, keybinding, behaviour)

### Backward compatibility

Not sure

### Example usage

```bash

```

### Additional context

_No response_

kbrdn1:
## Description

Multi-row selection in the TUI with a batch delete on top of it, and the non-interactive half of the same job on the CLI. Designed with the reporter in the issue thread: only `d` becomes bulk aware, the marks die on filter and manual refresh, the confirm names a count rather than listing rows, and `D` arms the branch deletion batch-wide.

Closes #484

## Type of change

- [ ] 🐛 Fix (bug fix)
- [x] 📝 Docs
- [x] ✅ Tests
- [ ] ⚡ Perf
- [ ] 🔧 Chore / CI / build

## Changes

- **`Space` marks a row, `d` deletes the marked set** (`toggle_select`). With nothing marked, `d` is the single-row delete it has always been. Only `d` reads the set, so the worktrees footer carries the count (` 3 of 12 · 2 marked `) and a `✓` column appears while the set is non-empty, both absent otherwise: a user who never presses `Space` keeps the pre-#484 table exactly.
- **Marks are keyed by on-disk path.** The fuzzy filter reranks indices on every keystroke, and worktree ids are only unique within one repo (workspace mode merges several). Opening `/` and the manual `f` clear the set; the background auto-refresh only prunes rows that vanished, since a 60s timer eating a selection mid-build would make the feature unusable. The main worktree cannot be marked, same reason `d` refuses it.
- **The confirm overlay snapshots its targets when it opens**, so an auto-refresh landing during the safety countdown cannot retarget the deletion. Today's path resolves through a filtered *index*, so a reorder mid-countdown moves the target; that is fixed here as a side effect. A batch reports `N selected` plus how many carry a branch instead of listing rows.
- **A batch never stops at the first error.** Every target is attempted through its own repo handle (a workspace batch can span repos, and nothing in the bulk path goes through the active-repo handle), the rows that failed stay marked for a one-keystroke retry, and the status line reads `removed 2 of 3 worktrees; failed: …`.
- **`gwm remove a b c`.** Every pattern is resolved before anything is touched, so an unknown or ambiguous one fails the whole command with nothing removed, which the `gwm list --format json | ... | xargs -n1 gwm remove` workaround from the issue thread could not do. Duplicate patterns collapse. A single pattern returns its error verbatim, so the pre-#484 command is byte-for-byte unchanged.
- **`cycle_sidebar_layout` moves from `Space` to `z`.** Space-to-mark is the lazygit / k9s / fzf convention, so the default is picked on merit; both pre-existing defaults are one `[tui.keys]` line away.

## Tests

- [x] `cargo test` passes locally
- [x] New tests added under `tests/` (TDD: a PR adding behaviour without tests is sent back)

New: 8 state-machine tests in `tui_app_tests.rs` (toggle, main-worktree refusal, target resolution and its fallback, stale-mark drop, prune across a refresh, clear on filter / manual refresh, the open-time snapshot, the batch status line), 2 render tests in `tui_table_render_tests.rs` (column absent, then glyph + footer count), 1 batch-confirm render test, 1 helper test, and 5 CLI tests (multi-pattern, resolve-first, per-pattern dry run, dedup, partial failure). Updated: the keymap / chord / footer suites for the `Space` and `z` defaults, and the delete-drain tests for the batch payload.

Both mark render tests close the sidebar first: its `State ✓ clean` row paints the same glyph, and with the sidebar open the positive test passed without a mark column existing at all.

## Screenshots / TUI captures

Not attached: the TUI half is validated locally before merge as usual.

## Checklist

- [x] README updated if CLI surface changed
- [ ] `examples/gwm.toml.example` updated if config schema changed — no schema change, and the example does not enumerate keymap defaults
- [x] No `unwrap()` on user-facing paths (return `GwmError` instead)
- [x] No `println!` in TUI render code (status bar only)

## Linked issues / docs

- Issue: #484
- Docs: `docs/2.tui/1.keybindings.md` gains a `bulk delete (Space + d)` section (EN + FR), `docs/3.cli/1.reference.md` documents `gwm remove <PATTERN>...`, and the gwm.toml / sidebar pages pick up `toggle_select` and `z`.

## Notes for reviewers

- **The `z` move is the one upgrade note.** `z` is now a shipped default, so a `.gwm.toml` binding a chord that *starts* with `z` (`top = ["z z"]` is the shape) becomes a prefix conflict and is refused at load time, with an error that does not mention #484. Called out in the changelog and both keybindings pages. Worth knowing: the issue thread itself suggested `z` as a free key for anyone who wanted `Space` back.
- **Stability policy.** `docs/6.development/3.stability.md` covers the `.gwm.toml` key *set* and disclaims TUI layout, but says nothing either way about shipped keymap default chords. Filed here as a question rather than changed in a feature PR: if default chords should be declared movable in a minor, that is a one-bullet edit to the policy.
- **Not in this PR, deliberately:** the TUI delete path writes no undo-journal entry and runs no `pre_remove` / `post_remove` hooks, unlike the CLI. Pre-existing, but a batch makes it bite harder (ten worktrees gone with no `gwm undo`). Follow-up issue rather than scope creep here.
- **Also not in this PR:** 14 internal doc anchors are already broken on `dev`. Verified my four new links resolve and that the count is identical before and after this branch.

https://claude.ai/code/session_0141pG1L6HnjzZsUchvnKBwY


kbrdn1:
## Codex review loop: 4 passes, stopped on convergence

| pass | findings | P0 | P1 | P2 | what they were about |
|:--|--:|--:|--:|--:|:--|
| 1 | 1 | 0 | 1 | 0 | the confirm overlay retargeting itself at the cursor row after a partial batch |
| 2 | 1 | 0 | 1 | 0 | removing by an id recycled during the countdown |
| 3 | 4 | 0 | 2 | 2 | the same window on the CLI batch, the double resolve in the pass-2 guard, the post-hook tally, workspace failures sharing an id |
| 4 | 2 | 0 | 1 | 0 (+1 P3) | the pass-3 guard again, plus four stale `Space` mentions |

Six findings fixed across `c76d91a`, `d48ecad`, `7fec3c2` and `7483697`, each with a test that goes red without its fix (verified by reverting the production change, not by reading the test).

**Two of them were real defects in the original design, not polish:**

- The confirm overlay recomputed its batch after a partial failure. `worktree::remove` prunes the admin entry before deleting the directory (#98), so a removal that fails on the filesystem drops its own row, the refresh prunes its mark, and the recomputation fell back to the **cursor row**. A second confirm deleted a worktree that had never been marked. The test drives two real worktrees through the drain; without the fix the retry batch targets the bystander.
- A worktree id is just the `.git/worktrees/<id>` entry name, and git hands it back to whoever recreates a worktree with that basename. The overlay snapshots its targets and fires after a countdown, and the CLI resolves the whole batch before running any hook, so both had a window where the id pointed elsewhere. Both now go through `worktree::remove_verified`, and the path check sits on the handle the prune acts on rather than in a wrapper that re-resolves the name.

### The one finding left on the table

> [P1] Refuse recreation at the same path too: if another shell removes and recreates the target **at the same path** during the countdown, git reuses the id, `wt.path() == expected` holds, and the replacement is destroyed. Validate a non-reusable generation or identity of the admin entry.

Not fixed, deliberately. Two reasons:

1. **The proposed remedy has nothing to build on.** A `.git/worktrees/<id>/` entry holds `gitdir`, `HEAD`, `commondir`, `index`, `logs/`, `COMMIT_EDITMSG`, `FETCH_HEAD`. No generation counter, no uuid, no creation stamp. Every approximation available changes under normal use: `HEAD` moves on any commit, `logs/HEAD` grows on any checkout, mtimes move on any write. A guard built on one of those refuses a legitimate delete whenever the user commits in another pane during the countdown, which trades a 4-conjunction scenario for a false positive on the everyday one.
2. **The consequence matches the instruction.** What the overlay showed, and what the user confirmed, is a path. Deleting the worktree that sits at that path is what was asked for. That is a materially different outcome from the finding this loop opened with, where the deletion landed on a path the user was never shown.

### Why the loop stopped here rather than at 8 passes

Signal 1 of the convergence rule fired on two consecutive passes: pass 3 flagged the guard written in pass 2, pass 4 flagged the same guard as amended in pass 3. Gravity followed the same curve, from "deletes a worktree the user never marked" down to "deletes the worktree at the path the user just confirmed for deletion". None of this code has been run against a real terminal yet, so the next useful information comes from the local TUI validation, not from a fifth pass.

**What is left to validate outside review:** the TUI half by hand (mark, unmark, `d` on a batch, `D` on a batch, the mark column and the footer count, `z` for the layout cycle), and the partial-failure path, which no test exercises end to end against a real filesystem error.

Issue #319, opened 2026-06-17
1.0: decide final shape of the exec/clean MVP cuts (#313)


kbrdn1:
## Problem

`gwm exec` / `gwm clean` (#313) shipped with deliberate MVP cuts. They are **additive** (not 1.0 blockers), but the final shape should be decided before the 1.0 compat pledge so later additions don't fight the initial design.

Deferred cuts:
- **`exec`**: bounded parallelism (a `--jobs`/`-j` flag? default sequential?) and the output model (current = sequential streamed stdio vs capture+interleave).
- **`clean`**: a `[clean]` config block (custom/extra artifact dirs beyond hardcoded `target/ node_modules/ dist/ build/`) and how it composes with the git-ignored + no-tracked-files safety gate.
- **both**: `--workspace` fan-out across repos (currently auto-rejected).

## Proposed solution

Decide each now — even if "defer, add additively post-1.0" — and record it so the 1.0 surface is intentional. Implement only what we want frozen at 1.0.

## Surface

cli / config

## Backward compatibility

Additive; the decision protects future additivity.

kbrdn1:
## Decision — exec/clean 1.0 surface

Resolved via #327. Verdict for every deferred cut: **defer additively** — none requires changing the 1.0 surface, because each can be added later without breaking the SemVer contract. The freeze is made intentional + tested rather than left implicit.

| Deferred cut | Verdict | Constraint protecting additivity | Follow-up |
|---|---|---|---|
| `exec` parallelism (`--jobs`) | defer | sequential stays the default through 1.x; `--jobs` opt-in | #324 |
| `exec` output model | defer | live streamed stdio is human-facing → not under SemVer (#318) | #324 / #325 |
| `clean` `[clean]` config | defer | new additive section; a profile **composes with — never bypasses** the safety gate | #324 |
| exec/clean `--workspace` fan-out | defer | refused today; refusal → success later is additive | #326 |
| exec/clean TUI surface | defer | TUI not under the SemVer pledge (#318) | #325 |

Lock: two regression tests in `tests/cli_binary.rs` pin the previously-untested `--workspace` refusal on both commands (failure + exit code 1 + message). Planned next features tracked in #324 (config with profiles), #325 (TUI surface), #326 (workspace fan-out).

PR #527, opened 2026-08-04
✨ feat(exec): container execution block on exec profiles

### LINKED ISSUE #421: [Feature]: container execution block on exec/aliases
## Problem

`lazyworktree` can run commands inside OCI containers, one isolated toolchain per worktree, no local runtime needed. `gwm` cannot.

## What the reference actually does

Read from the source (`internal/config/config_container.go`, `internal/multiplexer/container.go`), not only from `docs/guides/container-execution.md`. It is **not** a container runtime integration. It is a `container:` block attached to their existing custom commands:

```yaml
custom_commands:
  T:
    command: "go test ./..."
    container:
      image: "golang:1.22"
```

The schema is wider than the docs suggest. Nine fields, not four:

| field | behaviour |
|---|---|
| `image` | required; the whole block is discarded when it is empty |
| `runtime` | `docker` or `podman`; **auto-detection prefers podman** and falls back to docker |
| `interactive` | adds `-it` |
| `working_dir` | mount point inside the container, default `/workspace` |
| `mounts[]` | `source` (with `~` and env expansion), `target`, `read_only`, `options` (`z` / `Z` for SELinux) |
| `env{}` | extra variables, with `${VAR}` expansion |
| `extra_args[]` | pass-through to `docker run` / `podman run` |
| `args[]` | arguments placed after the image |
| the command | becomes `--entrypoint`, and a multi-word command is wrapped as `--entrypoint sh ... -c 'cmd "$@"' _` |

Three implementation details matter for what gwm should do differently, and all three are in the code rather than the docs:

1. `BuildContainerCommand` ends with `ShellQuote` over every argument and `strings.Join(args, " ")`. It produces a **shell string**, because the result is handed to tmux / zellij.
2. `WrapWindowCommandsForContainer` applies the same block to multiplexer windows, with `interactive` forced true. The `container:` block is not confined to custom commands.
3. Path environment variables are rewritten to the container path, and the rewrite is wrong: `MAIN_WORKTREE_PATH` is set to `working_dir`, the same value as `WORKTREE_PATH`, which claims the main checkout and the worktree are the same directory.

## The mount is the feature, and the obvious mount breaks git

This issue originally proposed:

```
docker run --rm -v <worktree>:/workspace -w /workspace <image> <cmd>
```

That form cannot run git. A linked worktree's `.git` is not a directory, it is a file holding an **absolute host path**:

```
$ cat /Users/.../worktrees/gwm-cli/fix-506-tui-table-bidi/.git
gitdir: /Users/.../gwm-cli/.git/worktrees/fix-506-tui-table-bidi/
```

That path does not exist inside the container. Reproduced on a throwaway repo by making the gitdir unreachable, which is exactly what mounting only the worktree does:

```
$ git -C $SB/wt rev-parse --short HEAD
fatal: not a git repository: (null)
```

So inside that container there is no `git status`, no `git describe` to stamp a version, no commit, no hook, and no coding agent that touches git at all. `go test ./...` passes; most of the rest does not.

**The reference has the same bug.** Confirmed in its code, not inferred from its docs: `BuildContainerCommand` mounts `worktreePath:workDir` plus user mounts and nothing else, and the docs state "The worktree path is automatically bind-mounted to the container's working directory. No configuration needed", with `mounts` examples covering only caches and credentials. Its largest documented use case is running a coding agent per worktree in a container, in which git does not answer.

For a worktree manager this is the most embarrassing possible gap, which makes it the one column here worth winning rather than matching.

### The fix, measured

Mounting the main checkout's `.git` at its **host-absolute path** is sufficient. Verified as far as writing, with the worktree tree relocated and the main checkout reduced to its `.git` alone:

```
$ ls $SB/main                     # only .git present
$ git -C $SB/workspace status --short --branch
## feat/x
$ git -C $SB/workspace commit -q --allow-empty -m "commit with only .git mounted"
$ git -C $SB/workspace log --oneline -1
ada5a3c commit with only .git mounted
```

Read, status and write all work. The main checkout's working tree is not needed, only its `.git`.

## Proposal

Since the mount has to reproduce a host-absolute path anyway, `/workspace` buys nothing. The default mirrors host paths:

```
docker run --rm \
  -v <worktree>:<worktree> \
  -v <main>/.git:<main>/.git \
  -w <worktree> <image> <cmd>
```

Two mounts, no path rewriting, and the container is byte-identical to the host as far as git is concerned. `working_dir` stays available for anyone who wants `/workspace`; taking it means owning the consequence, and the docs say so.

This remains a wrapper over an existing execution surface, not a new subsystem.

### Invariant: argv, never a shell string

The reference shell-quotes and joins because it feeds tmux. gwm executes argv directly (`gwm exec` forwards everything past `--` verbatim to a `Command`), so the container wrapper builds argv and never a string. Given GHSA-fffq-vg6f-gxqm was branch-name injection through hooks, this is stated as an invariant rather than left as a happy consequence.

### Environment

Mirroring host paths removes the remapping problem entirely: `GWM_PATH` and the `{path}` placeholder stay true on both sides. A wrapper that forwarded them into a `/workspace` container unremapped would ship the reference's `MAIN_WORKTREE_PATH` bug with more surface, since gwm exposes more of these variables (`docs/4.configuration/1.gwm-toml.md`).

### Entrypoint

Decide explicitly whether the user's command replaces the image entrypoint or runs as CMD, and write it down. Not a blocker: the images in play (`golang:1.22` and friends) have no entrypoint, so both forms work for them. The reference's `--entrypoint sh -c 'cmd "$@"' _` gymnastics exist because it builds a string, and do not carry over.

### Which surfaces the block attaches to

"The surface that already exists" covers more than it looks. Candidates: `gwm exec`, aliases, macros, lifecycle hooks, `gwm tmux` / `gwm zellij`, the PTY overlay, `gwm review`. The reference already wraps multiplexer windows, so this is not hypothetical.

One conflict has to be resolved rather than discovered: `interactive: true` allocates a TTY, while `gwm exec` is a sequential fleet command across N worktrees. A TTY per container across a fleet is meaningless. Either `interactive` is refused in fleet mode, or it is confined to single-worktree surfaces.

## Runtime support: nothing to integrate

Shelling out to `docker` / `podman` covers every runtime for free, because what matters is the **CLI**, not the engine behind it. Verified locally:

```
docker context ls
  orbstack *   OrbStack   unix:///Users/.../.orbstack/run/docker.sock
```

OrbStack is the active docker context and exposes a standard socket; the `docker` CLI talks to it without knowing. The same holds for Docker Desktop, Colima, Rancher Desktop, and native Linux Docker. Enumerating them as "supported runtimes" would be marketing, not integration. Documentation should say: works with anything exposing a Docker-compatible socket.

The OrbStack-specific surface (`orb` / `orbctl`, Linux machines) does not serve this use case and should not be instrumented.

Runtime auto-detection order is a decision, not a detail: the reference prefers podman. gwm should state its own order rather than inherit one by accident.

## Constraint

The 1.0 `exec` / `clean` surface is **frozen** (#319, `tests/contract_tests.rs`, defer-all additive policy). A `container:` block has to go through the additive path and respect that freeze.

## Priority

Sixth in the v1.7.0 queue: #484, #392, #420, #515, then this, then #422. It was originally filed as post-launch work with no observed demand. It moved into the line because the comparison page (#422) now sits behind it rather than ahead: container execution is a column in that table, and writing the table first would mean publishing a column gwm loses by default.

## Tests

`tests/exec_tests.rs`: argv construction for the container wrapper (pure, no spawn), and the interaction with the frozen surface.

Plus a mount test that is the point of this issue: for a **linked** worktree, the constructed argv contains the main checkout's gitdir mount. Without it the wrapper regresses to a container where git does not work, which is the failure this issue exists to avoid, and it is invisible to any test that only checks the worktree mount.

Actual container execution is not CI-testable and stays out.

kbrdn1:
## Description

`[exec.profiles.<name>.container]` runs a profile's command inside `docker run` / `podman run` instead of on the host. Three fields: `image` (required), `runtime` (auto-detected, docker first then podman) and `extra_args` (spliced in after gwm's own flags, before the image).

The wrapper is the cheap half. The mount is the feature: a linked worktree's `.git` is a **file** holding the absolute host path of `<main>/.git/worktrees/<id>`, so a container with only the worktree mounted answers no git command at all: no `git status`, no `git describe` to stamp a version, no commit, no hook, no coding agent that touches git. gwm mirrors host paths and mounts the main checkout's gitdir alongside, which is the column the reference implementation loses by default.

Closes #421

## Type of change

## Changes

- **`[exec.profiles.<name>.container]`** (`src/config.rs`): `image` required and non-empty, `runtime` optional, `extra_args` optional. `deny_unknown_fields`, so a key copy-pasted from the reference's wider schema is refused rather than ignored.
- **`build_container_argv`** (`src/exec.rs`): pure, no spawn, no `PATH` read, no filesystem. Emits `<runtime> run --rm -v <wt>:<wt> -v <main>/.git:<main>/.git -w <wt> <extra_args…> <image> <cmd…>`, skipping the second mount when the gitdir already lives inside the worktree (the main checkout, reachable via an explicit slug). `ContainerPlan` carries the resolved runtime + commondir across the fan-out; `resolve_container_runtime` takes an injected availability predicate so the tests never depend on the runner having a container runtime.
- **Wired into both exec surfaces** (`src/cli.rs`, `src/tui/app.rs`): the CLI fan-out (single-repo and `--workspace`, each repo resolving its own commondir) and the TUI exec picker. The same profile cannot mean "host" in one and "container" in the other.
- **`run_in_dirs_parallel` now takes an argv per item**: the mount is the worktree's own path, so no two worktrees run the same argv.
- **`validate_exec_profile` replaces `validate_exec_profile_command`** and destructures the profile exhaustively, so the next field added to `ExecProfile` cannot be silently skipped by a validator (the `gwm doctor` failure mode from #392).
- Per-worktree header announces a containerised run: `━━ feat-1 (/path) [docker rust:1.90]`.
- Docs EN + FR (`gwm-toml`, CLI reference, roadmap), README, `examples/gwm.toml.example`, CHANGELOG, ROADMAP.

## Decisions the issue asked to be resolved rather than discovered

- **Profile-only.** The inline `gwm exec -- <cmd>` is the frozen 1.0 surface (#319) and `ExecConfig` pledges it unchanged, so no config file gets to containerise a command line that used to run locally. `[aliases]` needed nothing despite the issue title: a gwm alias is argv substitution towards a subcommand, not a custom command, so `t = "exec --profile ci"` inherits the container through the profile it expands to.
- **argv, never a shell string.** Nothing is quoted or joined at any point. The reference joins + shell-quotes because it feeds tmux; gwm executes argv directly. Stated as an invariant and pinned by a test, given GHSA-fffq-vg6f-gxqm was branch-name injection through a shell hook.
- **No `interactive` / TTY knob.** `gwm exec` is a fan-out over N worktrees, where a TTY per container means nothing. Omitting it *is* the resolution of the conflict the issue names; it belongs to the surfaces that can honour it (multiplexer windows, PTY overlay).
- **Entrypoint**: the command is the container's CMD, so an image with an `ENTRYPOINT` receives it as arguments; `extra_args = ["--entrypoint", ""]` opts out. No `sh -c` gymnastics, those existed only because the reference builds a string.
- **Runtime order is stated, not inherited**: docker first, then podman (the reference prefers podman). An explicit `runtime` is honoured even when absent from `PATH`, since the spawn error names it better than a config error could.
- **Three fields, not nine.** `-w`, `-e` and `-v` are already `docker run` flags and `extra_args` lands last, so a repeat overrides gwm's own. No `~` / `$VAR` expansion; for caches the docs point at named volumes, which are portable where a host path in a committed `.gwm.toml` is not.
- **Environment**: `gwm exec` injects no `GWM_*` today, and mirroring host paths keeps `{path}` / `GWM_PATH` true on both sides, so the reference's `MAIN_WORKTREE_PATH` remapping bug has no counterpart here.

## Tests

- [x] `cargo test` passes locally (also on a stripped CI-like PATH: `PATH="$(dirname "$(command -v cargo)"):/usr/bin:/bin" cargo test`)

New coverage:

- `tests/exec_tests.rs`: argv shape, the **gitdir mount for a real linked worktree** (asserts the path the `.git` file actually names is covered by a mount, not merely that an extra `-v` exists), the dedupe (including a trailing-separator path, which is what git2 hands out for a main worktree), `extra_args` ordering, the no-quoting invariant, runtime resolution (preference, explicit override, no-runtime error), and the frozen inline surface staying host-only.
- `tests/config_tests.rs`: parse, `deny_unknown_fields`, missing `image`, empty `image` rejected on the config-validation path.
- `tests/cli_workspace_tests.rs`: end-to-end through the binary with `runtime = "echo"`, proving each repo mounts **its own** gitdir. Crossing them would be worse than the bug this fixes: git would answer, against the wrong repository.
- `tests/tui_app_tests.rs`: the picker wraps a container profile and leaves a plain one alone.

Each new guard was run **red** first by neutralising the line it protects (gitdir mount, dedupe, config validation, TUI wrap), then restored. A green guard proves nothing on its own.

Actual container execution is not CI-testable and stays out, per the issue.

## Notes for reviewers

- **Commit shape.** The first two commits are code+tests then docs: `run_in_dirs_parallel` changes signature, so src-alone and tests-alone both fail to build, and a commit that does not build breaks bisect. The four that follow are the Codex review loop, one per pass.
- **`extra_args` opens no escalation.** It is config-sourced, like `command`, and `command = ["docker", "run", "--privileged", …]` was already expressible in a profile before this PR. `gwm exec` deliberately has no trust gate (it runs the user's own command against their own worktrees), unchanged here.
- **`gwm doctor` does not probe the runtime binary** the way it probes hook binaries. That is deliberate and consistent with "an explicit `runtime` is honoured even when absent from `PATH`": a wrapper script or a `nerdctl` shim must not be reported as broken.
- The dedupe compares raw paths (`Path::starts_with` is component-wise) while the mount is normalised through `components()`. Both cases are pinned.

## Verified by running it, and what is NOT verified

The container path was exercised for real before merging, not only reasoned about: OrbStack 29.4.0, image `alpine/git`, a throwaway repo with a **linked** worktree.

```
# with the gitdir mount (what gwm builds)
$ git status --short --branch
## feat/x

# without it (the worktree alone)
fatal: not a git repository: (null)
```

That is the claim of #421, reproduced rather than deduced. The same run confirms the documented ENTRYPOINT behaviour: `alpine/git` declares `ENTRYPOINT ["git"]` and `command = ["status", "--short", "--branch"]` arrives as arguments to it.

**Four environments remain unvalidated. None of them is exercised by CI, and none of them can be, since container execution is not CI-testable (the issue says so):**

| environment | what is unproven there |
|---|---|
| **Linux + rootful Docker** | the `safe.directory` declaration is what keeps git answering as uid 0 against a host-owned tree. Verified by construction (the argv is pinned by a test), never observed. Also unobserved: files the command creates are owned by root on the host, which is documented but not measured |
| **SELinux enforcing** (Fedora, RHEL) | `selinux_relabel = true` appends `:z` to gwm's own mounts. The argv is pinned; the actual `EACCES`-without-it and the recursive relabel's cost on a large `.git` are not |
| **podman** | it is second in the detection order and accepted as an explicit `runtime`, but every real run so far went through docker |
| **overlay teardown on a long command** | `kill` runs `<runtime> rm -f <name>` and the unit test stands `touch` in for it. A container actually interrupted mid-build has not been observed |

A reviewer with any of these hosts is the fastest way to close the gap. Everything else in the PR is covered by the suite.

## Linked issues / docs

- Issue: #421
- Docs: `docs/4.configuration/1.gwm-toml.md` and `docs/fr/4.configuration/1.gwm-toml.md`


kbrdn1:
## Boucle de review Codex : arrêt sur stagnation après 4 passes

|---|---|---|---|---|---|
| 1 | 3 | 0 | 2 | 1 | code métier du diff |
| 2 | 3 | 0 | 1 | 2 | code métier du diff |
| 3 | 4 | 0 | 2 | 2 | 4/4 sur les correctifs de la passe 2 |
| 4 | 3 | 0 | 2 | 1 | 3/3 sur les correctifs des passes 2-3 |

**Basculement à la passe 3** : à partir de là, la review ne parle plus du wrapper mais du teardown écrit pour la passe 2, puis du nom écrit pour la passe 3. Signal d'auto-alimentation sur deux passes consécutives, donc arrêt.

**Trajectoire de gravité** : le finding le plus grave de la passe 1 était « git ne répond pas dans le conteneur sous Docker rootful » (le cœur de la feature). Celui de la passe 4 est « et si un pid réutilisé rencontrait un conteneur orphelin homonyme ».

**13 findings, tous traités** : 12 corrigés (safe.directory, TTY de l'overlay, refus Windows, teardown du conteneur, SELinux, refus du `:`, pid dans le nom, `--name` réservé, pré-résolution par repo puis workspace, cwd du teardown, teardown sur échec tardif du spawn), 1 réfuté par preuve : « vérifier l'identité avant `rm -f` » — le nom porte le pid, et deux processus vivants ne partagent pas un pid, donc un homonyme est forcément un orphelin dont la suppression est l'effet voulu.

**Ce que la review ne pouvait pas donner, et qui a été fait à la place** : le chemin conteneur n'avait jamais été exécuté. Une exécution réelle (OrbStack 29.4.0, `alpine/git`, worktree lié jetable) confirme la thèse de l'issue plutôt que de la déduire :

```
# avec le mount du gitdir (ce que gwm construit)
## feat/x
# sans
fatal: not a git repository: (null)
```

Elle valide aussi le comportement ENTRYPOINT documenté : `alpine/git` déclare `ENTRYPOINT ["git"]` et `command = ["status", "--short", "--branch"]` arrive bien en arguments.

**Reste à valider hors review**, non couvert ici : un hôte Linux avec Docker rootful (le chemin `safe.directory` + la propriété root des fichiers créés), un hôte SELinux enforcing (`selinux_relabel`), podman, et le teardown de l'overlay sur une commande longue réellement interrompue.

Issue #544, opened 2026-08-12
[Task]: act on ratatui maintainer's design feedback (capture width, empty space, borders, site contrast)


kbrdn1:
## Context

Unsolicited design feedback from [@joshka](https://github.com/joshka), ratatui's maintainer, on [ratatui/awesome-ratatui#403](https://github.com/ratatui/awesome-ratatui/pull/403). Worth treating as a source rather than an opinion: it is the person who maintains the library this TUI is built on, looking at the result with fresh eyes.

Three separate axes, filed together because they share a cause (how gwm presents itself visually) but they do not share a fix.

## 1. Captures are too wide, and the numbers are worse than they looked

> I always encourage TUI screenshots to aim a little narrower than you'd think (i.e. target 800-1000px wide, 1200 is about as wide as I'd want to go. This tends to work best in github README rendering etc, as you often want the TUI text and the surrounding text to be similar sized.

Measured across `docs/_capture/*.tape`:

| Declared width | Tapes |
|---:|---:|
| 1560 | 1 |
| 1500 | **13** |
| 1320 | 2 |
| 1280 | 2 |
| 1240 | 2 |
| 1200 | 2 |
| 1080 | 1 |

So the dominant width is **1500px**, 50% above his target range and 25% above the widest he would go. Only `demo.tape` and one other sit at the ceiling.

His reasoning is the operative part, and it is not aesthetic: at 1500px in a README, the TUI's text renders smaller than the prose around it, so the reader's eye has to change scale to read the screenshot. That applies to the docs site too, where every capture sits inside a column of body text.

## 2. Empty space, tables first

> Also a good rule is to try to cut down the amount of empty space in the screenshots - especially in tables.

This lands squarely on what #524 just shipped: several new captures are worktree tables, and a worktree table is mostly empty columns on a demo fixture with four rows. The fixture's row count and the terminal height are both ours to choose.

## 3. Bordered boxes

> I'm also a big fan of trying to avoid too many bordered boxes when you can by instead looking whether you can get away with backgrounds etc.

He attached a mock-up on the PR. This one is **not a docs change**: it is the TUI's own chrome. The sidebar alone stacks four bordered sections inside a bordered pane, and the worktrees table adds a fifth border around it. Worth prototyping against the real render before deciding, since borders also carry the focus indication today.

## 4. The site reads flat

> I noticed that it's fairly uniform brown and flat, could do with a bit of extra color pop in there / something to help the screenshots pop from the background a bit.

[...]

Issue #437, opened 2026-07-22
[Feature]: scroll the Working Tree pane from the Status context (Shift+J / Shift+K)


kbrdn1:
### Problem

Only the Recent Commits section scrolls today (`sidebar.scroll` / `max_scroll`). The Working Tree file-explorer tree (#300) is rendered at content height and clamped by the sidebar layout: on a large change set, entries beyond the pane height are simply unreachable from the keyboard.

### Proposed solution

When the Status pane has focus (context `status`), `Shift+J` / `Shift+K` scroll the Working Tree pane down / up:

- an independent scroll offset for the Working Tree section, clamped against a `max_scroll` republished by the renderer every frame (same owner split as the Recent Commits scroll state in `tui/state/sidebar.rs`)
- reset to 0 on worktree navigation, like the existing sidebar scroll reset
- two new rebindable actions (e.g. `wt_scroll_down` / `wt_scroll_up`) under `[tui.keys]`; `J` / `K` are free in the default keymap

### Alternatives considered

- Cycling focus across sidebar sections (a Tab-within-sidebar) and reusing `j` / `k` per section: heavier keymap and focus-model change for the same outcome. Deferred.

### Surface

tui (new view, keybinding, behaviour)

### Backward compatibility

No breaking change

### Example usage

```bash
gwm            # open the TUI
# 2 -> focus the Status pane
# J / K -> scroll the Working Tree file tree down / up
```

### Additional context

Related: #300 (Working Tree file-explorer pane). Pairs with the responsive sidebar min-height work: clamping section heights is only livable if the overflow scrolls.

kbrdn1:
Shipped via PR #452 (merged into dev, lands in v1.4.0). With the status pane focused, J/K (rebindable wt_scroll_down / wt_scroll_up) scroll the Working Tree file tree independently, clamped to the granted viewport; the offset resets on worktree navigation and on the commits/stashes toggle.

PR #546, opened 2026-08-12
✨ feat(tui)!: compact layout by default, bordered as the opt-out

### LINKED ISSUE #545: [Feature]: compact TUI mode, backgrounds instead of borders
## Problem

The TUI stacks five bordered boxes on a default screen: the worktrees pane, then Status, Issue/PR, Working Tree and Recent Commits inside the sidebar. Each border costs two columns and two rows, and the titles sit in the top rule, so a section carrying two lines of content occupies four.

Raised from the outside by [@joshka](https://github.com/joshka) in the review of [ratatui/awesome-ratatui#403](https://github.com/ratatui/awesome-ratatui/pull/403):

> I'm also a big fan of trying to avoid too many bordered boxes when you can by instead looking whether you can get away with backgrounds etc.

It also predates that feedback: a compact mode was on the wishlist early on and never got specified.

Parent issue, which groups the four axes of that review: #544. This one is the TUI half; the capture width and empty space live there.

## Proposal

A `[tui] compact = true` mode that replaces borders with background fills.

- **Sections lose their rules.** A one-line header on a slightly lighter fill (around 7% white on the default theme) delimits well enough. The keybinding leads the header (`1 WORKTREES`, `F ISSUE / PR`), and the counter moves to the right of that same line instead of taking a footer row.
- **Panes size to their content.** The worktrees pane currently reserves its full height whatever the row count, which is what makes a five-worktree demo look half empty.
- **The Status block folds.** `Branch` / `Created` / `Diff` / `State` are four short values on four lines today; they fit on one.

A rough HTML mock-up of the result exists and reads roughly twice as dense at the same screen height. Opt-in, default off: this changes what every existing screenshot looks like.

## What has to be resolved first

Three unknowns, and the first one is the actual design work rather than a detail:

1. **Focus indication.** Today the border colour says which pane is active (`border_style` on all three call sites). With no border, that signal has to move somewhere: the header accent, a lighter section fill, or a marker in the header. Removing the rules is the easy half; keeping "where am I" legible is the real one.
2. **Non-truecolor terminals.** A translucent white does not exist in ANSI 256. The fill needs an indexed colour that stays readable across the four shipped themes (`catppuccin`, `gruvbox`, `tokyo-night`, `claude-dark`), and the theme system is role-based, so this is likely a new role rather than a literal.
3. **Row selection already uses a fill** (`#3d3d3d` on the default theme). If sections gain one too, the two must stay distinguishable, including on a selected row inside a filled section.

## Surface

Three call sites carry `Borders::ALL` in `src/tui/ui.rs`:

- the list pane (worktrees / recent commits), with its title and counter footer;
- the sidebar sections, rounded (`BorderType::Rounded`);
- the overlay block (`overlay_block`), which is likely out of scope: a modal floating over content is exactly where a border earns its keep.

So the change is the first two, plus a `compact` key in `[tui]`, plus whatever the focus decision costs.

## Acceptance

- `[tui] compact = true` renders both panes and sidebar sections without rules, headers filled, panes sized to content
- focus stays unambiguous, verified on the four themes and on a terminal without truecolor
- default stays `false`, and the existing captures keep rendering as they do today
- a capture of the compact mode lands in the docs next to the default one, so the option is visible rather than buried in the schema

### LINKED ISSUE #549: [Feature]: modal titles belong in the border, not on two rows inside it
## Problem

Every modal spends four rows before its first line of content:

```
╭──────────────────────────────╮   ← border
│                              │   ← Padding::symmetric(2, 1)
│        New Worktree          │   ← overlay_title_lines, line 1
│                              │   ← overlay_title_lines, line 2 (spacer)
│  Type   : feat               │   ← the content finally starts
```

Two of those rows exist only to carry a title that ratatui can draw *in* the top rule for free. `src/tui/ui.rs`:

```rust
fn overlay_block(color: Color) -> Block<'static> {
  Block::default()
    .borders(Borders::ALL)
    .border_type(BorderType::Rounded)
    .padding(Padding::symmetric(2, 1))
    .border_style(Style::default().fg(color))
}

/// The detached modal title: a centred bold line in `color` followed by a
/// blank spacer row, prepended to a modal's content so the title sits
/// inside the rounded frame rather than embedded in the top border
/// (issue #217).
fn overlay_title_lines(title: &str, color: Color) -> Vec<Line<'static>> { … }
```

The detached form was a deliberate choice in #217. It is worth revisiting now that #545 has spent the whole codebase's chrome budget elsewhere: two rows per modal is the same order of saving as the two rows per section compact mode just recovered, and modals are the surfaces most likely to overflow a short terminal.

## Scope

- `overlay_title_lines` disappears; `overlay_block` takes the title and renders it in the top rule.
- 13 call sites build their content with `overlay_title_lines(...)` as the first element; they pass the title to the block instead.
- Height maths is `lines.len() + 2 /* border */ + 2 /* padding */` at each site. Those stay correct untouched: `lines` simply loses the two title rows.
- The bootstrap-report modal already renders a separate centred "Bootstrap Report" heading (`draw_report`); fold it the same way.

## Acceptance

- No modal renders a title row inside its frame; every one carries it in the top rule.
- Each modal is two rows shorter than before at the same content.
- The title keeps its accent + bold styling and stays legible against the rule.
- `tests/tui_modal_render_tests.rs` and `tests/tui_attach_modal_render_tests.rs` assert the title renders on the border row rather than on the first content row.

## Not in scope

Modal responsiveness — see the companion issue. This one is only about where the title sits.

kbrdn1:
## Description

The TUI stops drawing box rules around its panes and sidebar sections. Each is delimited by a filled one-line header instead, which buys back two rows and two columns per section, plus the row a counter used to take in a bottom rule. Modal titles move into the top rule in the same pass, two more rows per overlay.

**Compact is the default.** `[tui] layout = "bordered"` restores gwm's layout up to 1.7.

Closes #545, closes #549

## Type of change

> **Breaking**: `[tui] compact = true|false` never shipped, but if you tracked this branch, the key is now `[tui] layout = "compact"|"bordered"`.

## Changes

**Layout**
- `[tui] layout` — `compact` (default) or `bordered`. An unknown value is a hard load error, like the other `[tui]` enums.
- Sections lose their rules; the title keeps its bracketed keybinding and goes uppercase (`[1] WORKTREES`, `ISSUE / PR [F]`), the counter moves to the right of that same line.
- A `muted` rule marks the boundary between the two focusable panes — with every section delimited the same way, nothing else said where one ended.
- The worktrees pane sizes to its row count instead of reserving its share of the stacked split (the "five-worktree demo looks half empty" complaint). The share stays the ceiling; stacked only.
- Modal titles ride the top rule. All **15** overlays, enumerated from `View` + the `draw()` dispatch rather than by grepping for a symptom — that mistake cost a round, six modals titled six different ways.

**Focus**
- Moves from the border colour to the header: `focus`/`muted` on the text, `selection_bg`/`section_bg` on the fill.
- `[tui] dim_unfocused` (default **off**) additionally dims the inactive pane's body, in **either** layout. Off because it is a trade-off, not a strict win: that content is still readable information, and dimming costs contrast on a surface that is often a screenshot. Uses `DIM`, so semantic colours survive.

**Theme**
- New `section_bg` role for the header fill. Indexed rather than a translucent white, which has no ANSI 256 equivalent. Each preset takes the tone its own palette reserves for chrome bands; guarded against `selection_bg` on the default *and* all four presets — that invariant is also what makes the focused and unfocused header states distinct.

**Plumbing**
- `split_section_heights` takes the chrome cost as a parameter. The `+2` was hard-coded in the natural height and both floors; a floor left at 7 would have made the dense layout *taller*. At `chrome = 2` the values are identical, which the seven original tests check without changing their expectations.
- `Chrome` carries the cost and colours through the layout constraints, both scroll clamps, the inner width that trims a PR title, and the scrollbar rect. A flag re-tested at each site drifts, and drift is the #437 bug.

**Docs**
- Config pages, theme role tables, TUI pages — EN + FR. `examples/gwm.toml.example`. All 20 captures regenerated (the default changed, so every one of them illustrated a mode nobody has any more) and `bordered.tape` documents the opt-out.

## Tests

- [x] `cargo test` passes locally (96 suites)

Every guard was proved red by probing the production code, not by editing the test. Three started out passing vacuously and were reformulated until they could fail; a fourth was **deleted** rather than shipped green — the compact scroll clamp is not distinguishable from a correct one at the buffer level, and the arithmetic is pinned upstream on the pure solver instead. That is stated in the test that replaced it.

## Screenshots / TUI captures

Default (compact):

compact

`layout = "bordered"`, same geometry, same content:

bordered

The bordered shot fits 6 commits of 9 with a hole in the table; the compact one fits all 9, higher up.

## Checklist

- [ ] README updated if CLI surface changed — no CLI surface change

## Linked issues / docs

- Issues: #545, #549
- Parent: #544
- Follow-ups filed: #547 (Status block fold), #550 (modal responsiveness), #551 (rich PR/Issue view polish), #548 (pre-existing `cursor position` error on the overlay return path)

## Notes for reviewers

**What I did not verify, and will not claim.** The acceptance bullet reads "verified on the four themes and on a terminal without truecolor". What ran mechanically: the role is indexed / palette-real per preset, `section_bg != selection_bg` on the default and all four presets, zero box corners in compact, headers painted with the role, focus resolving on both panes. The visual pass across four themes and a non-truecolor terminal is the manual validation step, and it is where most of this PR's corrections came from.

**Scope grew, deliberately.** #545 called the overlay block "likely out of scope". It came back in as #549 during validation, in this PR rather than a stacked one: both touch `src/tui/ui.rs`, and the repo's own rule is to sequence rather than create avoidable conflicts on shared files.

**Three findings only a human or a picture could catch.**
1. gruvbox's first `section_bg` was `bg0` — the palette's own background, so the header was invisible on a gruvbox terminal while the `!= selection_bg` guard stayed green. A background role's contrast is against the *terminal*, which no unit test sees.
2. `Length` + `Percentage` do not add up to the body height; ratatui's default flex left the remainder as dead space *after* the sidebar, so the rows the pane gave back reached nobody. My guard measured where the sidebar *starts*, never where it ends.
3. Three call sites kept `+ 2 /* title */` after #549 moved it, leaving frames two rows too tall and scrollbars two rows too low.

**Codex review**: 3 findings in pass 1 (all fixed), 1 in pass 2 (finding 3 above, same diagnosis as the visual report), pass 3 running.


kbrdn1:
Follow-up for the deferred Status-block fold filed as #547.

PR #566, opened 2026-08-13
🔖 chore(release): v1.8.0, the density line + the capture pass behind it

### LINKED ISSUE #544: [Task]: act on ratatui maintainer's design feedback (capture width, empty space, borders, site contrast)

kbrdn1:
## Description

Cuts **v1.8.0**, and does the capture work that had to happen against the 1.8.0 binary rather than after it (the version chip is in every TUI capture).

Closes #544

## Type of change

- [x] 📝 Docs
- [x] 🔧 Chore / CI / build

## Changes

**The release** (`🔖 chore(release): v1.8.0`)

- `Cargo.toml` / `Cargo.lock` to 1.8.0, `[Unreleased]` migrated to `changelogs/1.8.0.md`, root index reset.
- Minor and not major on purpose: the published stability policy leaves *any visual detail of the ratatui interface* free to change in a minor. The frozen surfaces (CLI subcommands / flags / exit codes, `--format=json` schemas, daemon JSON-RPC, the `.gwm.toml` section set) are untouched. That reasoning is now written into the roadmap's current-state block, because a changed default invites the question.
- The four files a release commit bumps *around* without touching: `SECURITY.md` (1.8.x supported; a patch stays on its line, a new minor inverts that), the `Next up` preamble on all three roadmaps, the `Shipped highlights` row, and the `→` breadcrumb arrows left alone.

**The captures** (#544, axes 1 and 2, plus the contrast half of axis 4)

Every capture regenerated against the 1.8.0 build. Columns were **measured** with a `tput cols` tape at each candidate width, not estimated from pixels: 800px = 81 cols, 1000px = 103, 1500px = 159.

| | before | after |
|:--|:--|:--|
| full-TUI captures | 1500px | **1000px** (103 cols) |
| narrow layout | 1080px | **800px** (81 cols) |
| terminal background | `#2b2b2b` | **`#141414`** |

Four exceed the 800-1000 band, each because the surface does not fit below it, and each verified by looking:

- **`side-by-side` at 1400.** `SIDEBAR_MIN_WIDTH` is 120, so below it the sidebar stacks and the capture contradicts its page. At 121 it renders but `STATUS` reads `STA` / `cle`. It needs ~148 columns to be legible.
- **`?` overlay, Settings panel, `:` palette at 1240.** All three clamp their modal to a 64-column floor, so at 103 cols the modal is at its minimum and cuts the descriptions that are the point of the shot.
- **`gwm list` at 1340, `gwm agents` at 1220**, from their own output width (140 and 126 columns). Below that the shell wraps every row and scrolls the header and the `*` marker off the top.

Heights cut to content in the same pass (axis 2): the hero was carrying 12 empty rows, the narrow capture 21, the `o` overlay was ~85% empty. Measured per capture by finding the band between the last content row and the status bar; the two where a full-height rule defeats that measure were read by eye.

**The pipeline** (`🔧 chore(capture)`)

`generate.sh` was printing a tick over captures it had not written. vhs exits 0 either way, and the screenshot write races process exit: **measured, a bare tape landed its PNG one run in three**, and three of twenty-four silently kept their old asset on this cycle's first regeneration. Every tape carrying a `Screenshot` now has a `Sleep` after it (1/3 → 3/3 on the tape I measured), and the script asserts every declared target is newer than the run, retries once, and exits non-zero naming what did not land.

Proven red before being trusted green: against a tape pointing at a directory that does not exist, it detected, retried, and exited 1. It then caught a real miss on `bootstrap.tape` on its first real run.

`bootstrap.tape` is torn down at both ends as part of this. It is the one tape that mutates the fixture, and `gwm remove` keeps the branch on purpose, so a second run captured `branch already exists` instead of the report. The retry surfaced it by shipping exactly that PNG.

## Tests

- [x] `cargo test` passes locally: **97 `test result: ok`, 2948 tests, 0 failed**
- [x] `cargo clippy --all-targets -- -D warnings` passes (exit 0)
- [ ] New tests added under `tests/` — no behaviour change in this PR; it is a version bump, regenerated assets, and shell tooling. The pipeline guard was proven against a deliberate red case rather than by a Rust test, since it exercises vhs.

## Screenshots / TUI captures

Every PNG and GIF in the diff. The hero, the compact/bordered pair and `side-by-side` are the ones worth a look.

## Checklist

- [x] Branch follows the convention (`chore/release-1.8.0`, matching `chore/release-1.0.3`)
- [x] `CHANGELOG.md` updated — migrated to `changelogs/1.8.0.md`, root reset to a bare `[Unreleased]`
- [x] README updated (the changed default is user-visible)
- [ ] `examples/gwm.toml.example` — no schema change in this PR

## Linked issues / docs

- Issue: #544
- Release notes: `changelogs/1.8.0.md`

## Notes for reviewers

Two things left honest rather than dressed up, both stated in the commit bodies:

1. **The palette still cuts its two longest action descriptions.** Reaching them means 1500px, which undoes the pass for one capture. 129 columns is what a user actually sees.
2. **The `PATH` column is hard-clipped with no ellipsis.** Deliberate and documented at the call site: it is `Fill(1)`, outside `trunc`'s funnel by design, and it vanishes last. The old 1500px hero clipped it too, just later. Not introduced here.

Axis 4 of #544 (the docs site reading flat) is half-addressed: the darker terminal makes the frames sit *on* the page instead of *in* it, but the page's own palette lives in `kbrdn1/kbrdn-docs`.


kbrdn1:
Two additions since the description was written, both pushed:

**The `github-linking` capture.** It is the one shot outside `generate.sh`'s loop (the Issue/PR pane needs a live remote, the demo fixture has none), and it was still filtering for a `docs/#206` worktree deleted several releases ago. The filter matched nothing, so the image was frozen at 1500px on the old background with a `gwm 1.7.1` chip: the only capture in the docs not following the rest.

The filter is gone rather than repointed at another number that would rot the same way. The pane follows the selected row, so the tape works from any branch with an open PR. Here that is this one, and the shot carries a real detected PR with its CI rollup mid-run. Its caption promised an auto-linked issue the pane does not render, in both languages, and now says what is in the image.

**The repo banner** (`docs/_assets/promo.png` + `promo-light.png`) reads v1.8.0, with the test count the suite actually reports and a branch graph carrying five branches from this cut. Each label is a real prefix cut on a hyphen, not a rewrite: the kit clips past roughly 22 characters without saying so, and `refactor/#550-modal-responsiveness` and `fix/#548-clear-without-cursor-query` would both have lost their tail.

Both are recorded in `changelogs/1.8.0.md`.

PR #574, opened 2026-08-14
♻️ refactor(cli,tui): a message gwm prints reads without an em dash

### LINKED ISSUE #567: [Task]: the em dash sweep stopped at the docs, and the binary prints 135 of them
## Scope

`#516` swept `docs/`, and `#543` finished `skills/SKILL.md`. Neither touched `src/`, where **135 string literals contain an em dash**. Most of them are printed to the user.

```
$ grep -rn '"[^"]*—[^"]*"' src/ --include='*.rs' | wc -l
135
```

A sample, all of them error messages a user reads on stderr:

```
exec: --profile and an inline `-- <cmd>` are mutually exclusive — the profile carries the command
exec: profile `{profile}` has an empty `command` — give it an argv array like …
exec: no container runtime found on PATH (looked for {}) — install one or set `runtime` in …
exec: `[container]` is not supported on Windows — the wrapper mirrors host paths, …
```

And at least one in the TUI chrome itself, visible in every capture of the fuzzy filter:

```
[/ filter — type to narrow · enter confirms · esc clears]
```

## Motivation

The rule this repo writes under is that published prose does not use the em dash, and the binary's own output is as published as the README. It reads inconsistently today: a user who hits an `exec` error sees a dash the documentation for that same feature no longer uses.

There is a second, smaller reason specific to the terminal. The em dash is a wide glyph in some fonts and a narrow one in others, and the width work in this line (#554 / #560 / #562 / #563) made every truncator measure in terminal cells precisely because that class of character is where the measure used to be wrong. Fewer of them in strings gwm controls is one less place for that to matter.

## Kind

refactor (no behaviour change)

## Notes

**Not a `sed`.** #516's write-up is explicit that an automated substitution produced prose like `Source: …/gwm-cli: latest stable: … : MSRV 1.95`, three colons in one sentence, and four lines had to be reworked by hand. The connector depends on what the dash was standing in for: a colon where it introduces an explanation, a comma for an apposition, a semicolon where the clause already carries a colon.

**Two categories that must not be rewritten**, same rule as #516:

- doc comments quoting a spec or a command's real output;
- any string a test pins. `grep -rn '—' tests/` before touching a literal, or the sweep turns green tests red for no behavioural reason.

**A guard is worth more than the sweep.** Without one this comes back the next time someone writes an error message. A test asserting that no string literal under `src/` contains `U+2014` would hold the line, with an explicit allow-list if any legitimate case survives.

kbrdn1:
## Description

The rule this project writes under is that published prose does not use the em
dash, and the binary's own output is as published as the README. #516 swept
`docs/` and #543 finished the skills; neither reached `src/`. Two surfaces, not
one: 165 dashes across 161 string literals, and 49 more in `gwm --help`, which
`clap` builds out of the doc comments on the CLI types. Plus a test on each so
the habit does not come back.

Closes #567

## Type of change

## Changes

- **The literal guard, committed first and red on its own** (`5385fd9`). It
  listed 161 offenders as an assertion failure, which is the contract the next
  commit satisfies. A grep cannot state it: `grep -rn '"[^"]*—[^"]*"'` finds
  138 lines where the real figure is 161 literals, it cannot see one that spans
  lines, and it reads the quotes inside a doc comment as a literal's delimiters.
  That last one decides the design, since the ~1900 em dashes in comments are
  deliberately out of scope.
- **The literal sweep**, one connector chosen per call site rather than
  substituted. A colon where the dash introduced the remedy, a semicolon where
  the clause already carried a colon, a full stop where a second colon would
  have made the sentence unreadable, a comma for an apposition. #516 is why this
  is not a `sed`: an automated pass there produced `Source: …: MSRV 1.95` and
  four lines had to be reworked by hand.
- **The TUI separators**, where the dash was not punctuation but a value.
  `/ filter`, the create form hints and the loader detail take `·`, which the
  segments beside them already used. Where the tail was already `·`-joined a
  colon separates the title from the list, so the two roles stay distinct. Two
  labels became parenthetical, which is what they always were: `(conflict)` and
  `New Worktree (free-form)`.
- **The help guard** (`bbe5561`, also red on its own, 174 lines). Excluding doc
  comments is only sound while a doc comment stays a comment, and clap's derive
  makes those in `src/cli.rs` the published output. Rather than teach the
  scanner which ones clap consumes, which is a proxy that would still miss
  `long_about`, `after_help` and whatever clap renders next, this walks the help
  tree and reads what a user reads.
- **The help sweep**, with two shapes the messages did not have. A dash used in
  **pairs** to bracket an aside became a parenthesis, since commas would collide
  with the ones already inside it (`gwm review`, `gwm statusline`). In the
  `gwm pr` placeholder table the dash was a **column separator**, the case #516
  found in the `docs/` tables, so the row now reads `=`. `gwm completions zsh`
  is fixed by the same edit, 23 dashes to none.
- **Two assertions** pinned a rewritten string and travel with it, so the suite
  is green at every point of the history.

## Tests

- [x] `cargo test` passes locally (2966 tests, 100 binaries, 0 failures)
- [x] New tests added under `tests/` (`tests/em_dash_guard_tests.rs`)

Four tests, because a scanner that desynchronises reports an empty offender list
exactly like a clean tree does: the literal sweep with a floor under the 4867
literals `src/` holds today, the help walk with a floor on pages visited, a
fixture pass exercising each construct in both directions (raw strings, escaped
quotes, nested block comments, the lifetime tick), and a pass reading the two
hardest constructs off the real tree rather than off a fixture.

## Screenshots / TUI captures

Deliberately not regenerated here. Several captures show a rewritten string
(`filter`, `keybindings`, `palette`, `countdown`, `shell-init`, and `demo.gif`
walks two of those states), and each one needs an eyeball that cannot be
automated. `docs-sync.yml` fires on a push to `main` and never on `dev`, so the
published site keeps the current captures until the next release cut. Tracked in
#575 rather than stacked here, per the house rule on scope.

## Checklist

- [ ] README updated if CLI surface changed (no surface change: same commands,
      same flags, same exit codes; only the wording of the help text)
- [ ] `examples/gwm.toml.example` updated if config schema changed (unchanged)

## Linked issues / docs

- Issue: #567
- Follow-up: #575 (doc captures)
- Prior sweeps: #516 (`docs/`), #543 (skills)

## Notes for reviewers

Three points where I read the issue against the tree rather than following it:

1. **The issue undercounts twice, in the same direction.** Its `grep` reports
   135 string literals; a scanner finds 161, and the help surface it does not
   look at carries 49 more. Both numbers are in the commits that fix them.
2. The issue says a string a test pins must not be rewritten. CLAUDE.md's own
   exception list says the opposite, that a pinned string has its assertion
   updated, and that is what happened here. It was two assertions, not the churn
   the issue anticipated: most of the 101 em-dash literals under `tests/` are
   assert *messages*, which are read by whoever runs the suite and are not
   published.
3. Nothing here touches a frozen contract. `json_api.rs` and `contract.rs` carry
   em dashes only in doc comments, and the one JSON-adjacent literal
   (`daemon.rs`, a Windows pipe-bind failure) is an error message, not a pinned
   field. Checked before the sweep, not after.

Also measured and left alone: the en dash appears three times in `src/`, all in
comments, all of them the range in `3–5×`. It is a numeral, not a connector, so
both guards cover U+2014 only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and #543 finished the skills; neither reached `src/`, so a user hitting an
   `exec` error read a dash the documentation for that same feature no longer
   used. 165 of them across 161 string literals, in every error message, status
-  line and TUI hint the binary carries.
+  line and TUI hint the binary carries, and 49 more in `gwm --help`, which
+  `clap` builds out of the doc comments on the CLI types. The completion
+  scripts carried the same text and are fixed with it: `gwm completions zsh`
+  went from 23 to none.
 
   The connector was chosen at each call site rather than substituted. A colon
   where the dash introduced the remedy, a semicolon where the clause already
@@ -29,11 +32,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   create form hints and the loader detail now read the way the segments beside
   them already did.
 
-  Comments and doc comments are untouched, roughly 1900 of them: a doc comment
-  quoting a spec or a command's real output has to stay verbatim. A test
-  enforces the rule from here on, scanning `src/` with a Rust literal scanner
-  rather than a grep, since telling a literal from a doc comment is the whole
-  difference between a guard that holds and one nobody can keep green.
+  Comments are untouched, roughly 1900 of them: a doc comment quoting a spec
+  or a command's real output has to stay verbatim, and it is read in the source
+  rather than printed. Two tests hold the rule from here on. One scans `src/`
+  with a Rust literal scanner rather than a grep, since telling a literal from
+  a doc comment is the whole difference between a guard that holds and one
+  nobody can keep green. The other walks `gwm --help` and every subcommand's,
+  because that is where a doc comment stops being a comment, and no scanner
+  can answer that question as reliably as reading what the binary prints.
 
   Several doc captures still show the old wording and are tracked separately.
   The published site follows `main`, never `dev`, so they go stale at the next
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **A message gwm prints reads without an em dash**
+  ([#567](https://github.com/kbrdn1/gwm-cli/issues/567)). The rule this
+  project writes under is that published prose does not use the em dash, and
+  the binary's own output is as published as the README. #516 swept `docs/`
+  and #543 finished the skills; neither reached `src/`, so a user hitting an
+  `exec` error read a dash the documentation for that same feature no longer
+  used. 165 of them across 161 string literals, in every error message, status
+  line and TUI hint the binary carries.
+
+  The connector was chosen at each call site rather than substituted. A colon
+  where the dash introduced the remedy, a semicolon where the clause already
+  carried a colon, a full stop where a second colon would have made the
+  sentence unreadable. In the TUI the dash was often not punctuation at all
+  but a separator, and there is already one in use there, so `/ filter`, the
+  create form hints and the loader detail now read the way the segments beside
+  them already did.
+
+  Comments and doc comments are untouched, roughly 1900 of them: a doc comment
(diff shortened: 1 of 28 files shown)


kbrdn1:
Follow-up for the doc captures: #575.

Issue #581, opened 2026-08-14
[Bug]: doc captures are rendered at 1x and blur on the published site


kbrdn1:
## Description

Every capture under `docs/**/_assets/` is rendered by vhs at 1x: the tapes set
`Set FontSize 15` with a pixel `Set Width` between 800 and 1400, so the PNG that
ships is exactly as wide as the terminal it photographed. On
[gwm.kbrdn.dev](https://gwm.kbrdn.dev) those images are painted wider than they
are, and on any HiDPI display they are painted at twice the CSS width again, so
the browser upscales and the text goes soft.

Measured on the deployed page (`/tui/`, Chromium):

| viewport | DPR | content column | `hero` intrinsic | painted | physical px needed |
|---|---|---|---|---|---|
| 948 | 2 | 898 | 1000 | 898 | 1796 |
| 2560 | 1 | 1230 | 1000 | 1230 | 1230 |

`hero.png` is 1000 px wide and the site paints it at 1230 px even at DPR 1, so
the wider the screen, the worse it reads. `srcset` is empty, so there is no
higher density source to pick from either.

## Steps to reproduce

1. open https://gwm.kbrdn.dev/tui/ on a HiDPI display
2. look at the worktree table text in the hero capture
3. widen the window: the capture keeps growing past its 1000 px source

## Expected behaviour

Captures read as crisply on the site as the TUI does in a terminal.

## Actual behaviour

Text edges are soft at every viewport on a HiDPI display, and soft on every
display once the window is wide enough for the content column to exceed
1000 px.

## Severity

🟢 Cosmetic (visual only)

## Notes

vhs 0.11.0 has no `Set Scale` (`Unknown setting: Scale`). Density comes from
doubling `Set FontSize`, `Set Width`, `Set Height` and `Set Padding` together:
the terminal grid stays put (measured 103x31 columns before, 104x32 after) and
the PNG comes out at twice the pixels. `hero.png` goes from 272K to 641K that
way.

Two constraints this has to respect:

- **The published crate is at 9.8 MiB of the 10 MiB crates.io limit**
  (`cargo package --list` on 1.8.0: 442 files, 14.7 MiB, 9.8 MiB compressed).
  `docs/` alone is 9.2 MiB of the tree and `Cargo.toml` has no `exclude`, so
  doubling the captures makes `cargo publish` fail. The captures have no
  business in the published crate anyway.
- **`generate.sh` deliberately skips `demo.tape` and `github-linking.tape`**,
[...]

Issue #515, opened 2026-08-04
[Feature]: per-worktree notes, editable and visible from the TUI


kbrdn1:
## Problem

A worktree is a task in flight, but gwm holds none of the context that goes with it. It knows the branch, the linked issue, the diff against base and the agent session. It does not know where you were: what you had just figured out, what is blocking, what to check before opening the PR.

With eight worktrees open, that context lives in your head or in a scratch file somewhere outside gwm. Coming back to a worktree after two days means reconstructing it from `git log` and memory. The TUI shows everything about a worktree except the one thing only you can write down.

## Proposed solution

A note attached to a worktree, plain markdown, editable from the TUI and readable from the CLI.

Four things ship in the first cut, and nothing else:

1. **A storage that is decided, not left open** (below), with the lifecycle rule stated: the note lives as long as the branch.
2. **Edit from the current row.** `N` opens the note in `$EDITOR`. The handoff already exists: `editor_cmd` in `.gwm.toml`, then `$EDITOR`, then `vi` (`src/tui/app.rs:5901`). `N` is free in every keymap context. `i`, which is what the reference implementation uses, is not: it is `LinkPrompt`.
3. **A binary marker in the worktrees table.** This row carries a note, or it does not. Deliberately not an icon, not a colour, not a badge; see Out of scope.
4. **`gwm note show [slug]`, and a `note` field on `JsonWorktree`.** gwm is a CLI with a TUI on top, not the reverse. Every other capability here has a machine surface, and a note that exists only inside the TUI is off-contract for a project shipping a frozen JSON schema, a daemon, `subscribe` and a statusline. The field is additive, so it lands under the same `SCHEMA_VERSION`.

## Storage

There are four options, not two, and the two originally listed here were the weakest of them.

| | survives `gwm remove` | readable from the main checkout | plain text on disk | orphans |
|---|---|---|---|---|
| in the worktree (`.git/gwm-note`) | no | no | yes | none |
| in the gwm config dir | yes | yes | yes | yes, needs a sweeper |
| `git config branch.<name>.gwm-note` | yes | yes | **no** | none |
| **`.git/gwm/notes/<branch>.md` in the main checkout** | **yes** | **yes** | **yes** | **none in practice** |

[...]

Issue #408, opened 2026-07-21
[Tracking]: Agent session pane


kbrdn1:
## Context

Agent session tracking is the single feature gap that `lazyworktree` covers and `gwm` does not, and it is the one with the strongest 2026 pull ("one isolated worktree per AI agent"). It is the launch hook: it should ship before the wider launch push, not after.

Reference implementation read: `chmouel/lazyworktree`, `internal/app/services/agent_{processes,sessions,registry,watch}.go`. Two limits worth beating rather than copying:

- it gives up on Windows entirely (`agent_processes.go:55`, `Refresh` returns `nil` when `runtime.GOOS == "windows"`)
- it hardcodes Claude and pi (`classifyAgentProcess` only matches `claude` / `claude-code` / `claude desktop` / `pi`)

`gwm` ships on Scoop and winget, so silently dropping Windows is not an option here.

## Approach

Detect sessions from on-disk artefacts rather than by scanning processes. Every agent CLI persists its session state, and that state carries the working directory. This is `std::fs` only: no OS API, no `ps` / `tasklist`, testable against a `TempDir`.

Four storage schemes were inspected on a real machine, and they are mutually incompatible, which is what justifies a trait rather than a single configurable path:

| Agent | Location | How the cwd is recovered |
|---|---|---|
| Claude Code | `~/.claude/projects/<slugified-cwd>/*.jsonl` | the directory name encodes the cwd |
| Codex | `~/.codex/sessions/YYYY/MM/DD/rollout-*.jsonl` | first line, `session_meta.payload.cwd` |
| opencode | `~/.local/share/opencode/storage/project/<hash>.json` | `worktree` field |
| Mistral Vibe | `~/.vibe/logs/session/session_<ts>_<id>/meta.json` | `environment.working_directory` |

## Sub-issues

- [ ] Spike: pin down the Claude Code cwd-slug convention
- [ ] `SessionSource` trait + four backends + activity detection
- [ ] Generic detail overlay (shared with the rich PR/Issue view)
- [ ] TUI surface: agent column, Status pane detail, `a` keybinding
- [ ] Daemon + statusline exposure
- [ ] Process-level liveness on macOS/Linux (deferred, post-demo)

## Out of scope

Container execution (tracked separately) and the rich PR/Issue view (separate priority, but shares the overlay).

kbrdn1:
Sub-issues filed:

- #409 spike: Claude Code cwd-slug convention (blocks the Claude backend only)
[...]

PR #600, opened 2026-08-25
🐛 fix(tui): the note column captions itself (#595)

### LINKED ISSUE #595: [Fix]: the note column has no header, so it reads as part of I/P
## Problem

The note column (#515) is already conditional: it exists only when a visible row carries a note (`tui/ui.rs:849`, `let show_note = visible.iter().any(|w| w.has_note)`). Its **header** is not a caption though, it is empty:

```rust
header_cells.push(Cell::from("I/P"));
if show_note {
  header_cells.push(Cell::from(""));   // <- the note column's header
}
header_cells.push(Cell::from("NAME"));
```

So the `≡` markers sit under a blank header, immediately right of `I/P`, and read as a third slot of the `I/P` group rather than as their own column. The original reasoning was "caption-less: the marker is binary" (`tui/ui.rs:846-849`), which explains the empty string but not the ambiguity it creates next to a two-slot column.

## Proposed solution

Give the note column its own header, still shown only when the column is (that part already works):

- a caption that separates it visually from `I/P`: `≡`, `NOTE`, or a single-cell glyph matching the marker,
- the same `muted` + `BOLD` header style as the rest of the row.

The narrow-terminal argument that made the column conditional still holds, so nothing changes for a user with no notes.

## Surface

tui (rendering)

## Backward compatibility

No breaking change: the column already appears and disappears; only its caption changes.

## Additional context

- The mark column (#484) and the AGENT column (#408) follow the same appear-when-non-empty rule; AGENT has a caption, the mark column does not. This makes the note column consistent with AGENT.
- Header captions are asserted in `tests/tui_table_render_tests.rs`; a caption that only appears with notes needs both branches pinned, and the "no notes" branch is the one that would pass vacuously if the assertion only looked for presence.

kbrdn1:
## Description

The note column (#515) shipped caption-less on the grounds that its marker is binary. That is true and it does not answer the question the layout asks: an empty header immediately right of the two-slot `I/P` group put the marker under blank space next to that group, so it read as a third `I/P` slot rather than as a column of its own.

Header and marker now share one glyph, and both move off `≡` onto `nf-oct-markdown`, the glyph the Working Tree pane already paints on a `.md` file, because a gwm note is one. The conditional rule is untouched: a user who never writes a note keeps the same table, caption included.

Closes #595

## Type of change

- [x] 🐛 Fix (bug fix)
- [ ] 📝 Docs
- [ ] ✅ Tests
- [ ] ⚡ Perf
- [ ] 🔧 Chore / CI / build

## Changes

- `NOTE_ICON` (`nf-oct-markdown` + its trailing space) replaces the empty header cell and the `≡` marker; header and rows paint the same glyph, header in `muted` + `BOLD` from the row style, rows in the neutral `name` role as before.
- The column goes from `Constraint::Length(1)` to `Length(2)`: the glyph plus the trailing space this tree gives every nerd-font glyph, since most render two cells wide and the space is what the second cell eats instead of the neighbouring column.
- New render test pinning the caption **by position**, plus the two existing marker tests moved onto data rows only.
- Docs EN + FR and `skills/SKILL.md` stop pasting a literal `≡`.

## Tests

- [x] `cargo test` passes locally (exit 0, 103 `test result: ok`, 0 failures)

`the_note_column_captions_itself_between_ip_and_name` asserts the glyph's **index** in the header line sits between `I/P` and `NAME`. A `contains` on the flattened buffer would have been satisfied by the row marker and stayed green with no header at all, which is the failure mode this column already had. For the same reason `a_row_with_a_note_shows_the_marker` and `the_note_marker_is_only_on_the_rows_that_carry_one` now read data rows only: the header carries the glyph too, so a whole-buffer count passes with the marker gone.

Test helpers gained `draw_once_lines` / `is_header` so an assertion can tell a header cell from a row cell, which a flattened buffer cannot.

## Screenshots / TUI captures

Captured with `vhs` against the `acme-api` demo fixture, using the binary built from this branch. The header reads `I/P <glyph> NAME`, `fix-57-rate-limit-headers` carries the marker, and `NAME` stays aligned across every row. Attached in the review thread rather than committed, since it is not a published asset.

## Checklist

## Linked issues / docs

- Issue: #595
- Related: #515 (the column), #484 / #408 (the same appear-when-non-empty rule)

## Notes for reviewers

**The glyph is not pasted into `docs/`.** No page under `docs/` carries a Private Use Area codepoint today and the published site ships no nerd font, so a literal glyph would render as a tofu box for every reader. The prose names what it is and where the reader has already seen it instead.

**`docs/_capture/demo.gif` is now stale.** The recording writes a note and the table then shows the old `≡` for about ten seconds. It is deliberately out of this PR: `demo.tape` is not in `generate.sh`'s loop, the GIF is a binary that would conflict with #581, and the house rule prefers a follow-up over scope creep. Follow-up issue filed.

**Width, honestly.** The second cell is padding, not content, on a terminal that renders the glyph single-width, which is what the capture above shows. It follows the convention the Working Tree pane already uses rather than betting on the terminal, and it costs one cell versus the previous `Length(1)`, only while a visible row carries a note.


kbrdn1:
Follow-up for the stale recording: #601.

Issue #589, opened 2026-08-25
[Feature]: a setting for which direction a mux pane opens in


kbrdn1:
## Problem

`SpawnMode::Split` is hard-coded at every TUI call site and carries no direction:

- `tui/app.rs:5165-5167` (`open_in_mux_pane`, the `t` key) builds `SpawnMode::Split` unconditionally.
- `tui/mod.rs:1417-1420` does the same for a `open_in = "mux_pane"` macro.
- `cli.rs:3971` is the only place a user chooses, through `--split` (`Window` otherwise).

And `Split` itself is directionless: `build_tmux_command` emits `tmux split-window -c <path>` (no `-h` / `-v`, so tmux picks its default vertical stack) and `build_zellij_command` emits `zellij action new-pane --cwd <path>` (no `--direction`). On a wide screen the resulting stacked pane is the wrong half of the split, and there is no way to say so.

## Proposed solution

A `[tui]` setting for how a mux pane opens, read by the two TUI call sites and overridable per invocation on the CLI:

```toml
[tui]
# where a new mux pane goes: "right" | "down" | "window"
mux_pane_direction = "right"
```

- `right` / `down` map to `tmux split-window -h` / `-v`, `zellij action new-pane --direction right|down`, and (if #herdr lands) `herdr pane split --direction right|down`.
- `window` keeps today's `SpawnMode::Window` behaviour (new tmux window / zellij tab) for people who want the full screen.

`SpawnMode` grows from two variants to three, or keeps two and gains a direction field. The second shape is probably cleaner: `Split(Direction)` and `Window`.

Also worth exposing in the Settings panel TUI tab, like the other `[tui]` knobs (`src/tui/state/config_panel.rs`, `SettingsTab::Tui.fields()`).

## Surface

config (.gwm.toml schema) + tui

## Backward compatibility

No breaking change if the default reproduces today's behaviour. Today tmux stacks vertically by default, so the strictly compatible default is `down`; `right` is the better default on a wide screen and is a visible behaviour change. Pick one deliberately and say so in the changelog.

## Additional context

- `gwm tmux --split` / `gwm zellij --split` should take the same direction, either from the config or from a new flag.
- The argv builders are pure functions with pinned tests (`tests/multiplexer_tests.rs`), so each direction is one more assertion rather than a new test harness.

kbrdn1:
Delivered in #606, merged into `dev`.

[...]

Issue #36, opened 2026-05-19
[Feature]: Multi-repo mode — list worktrees across a workspace root


kbrdn1:
## Problem

\`gwm\` operates on a single git repository. Many developers organise their work as \`~/Projects/<repo1>/\`, \`~/Projects/<repo2>/\`, ... and routinely want a bird's-eye view of \"which worktrees do I have open across all my projects?\" Today that requires \`for d in ~/Projects/*; do (cd \"\$d\" && gwm list); done\` — usable but clunky, and the TUI is single-repo only.

## Proposed solution

Add a workspace mode triggered by:

- \`gwm --workspace <dir>\` (e.g. \`gwm --workspace ~/Projects\`), or
- Auto-detect: if the CWD is not a git repo but it contains child directories that are, prompt: \"No git repo here. Open <dir> as a workspace? [Y/n]\".

In workspace mode:
- The TUI table gets a \`REPO\` column at the front, listing every worktree of every direct-child git repo.
- Selection-driven actions still apply (\`l\`, \`o\`, \`d\`, \`b\`, \`Enter\`, etc.); the repo context is resolved per-row.
- \`gwm.toml\` is per-repo as today (no workspace-level config in v1 of this feature) — each row inherits its own repo's config.

CLI surface:
- \`gwm list --workspace ~/Projects\` outputs the merged table.
- \`gwm create\` in workspace mode requires an explicit \`--repo <name>\` to disambiguate.

## Alternatives considered

- A shell loop over \`gwm list\`: works but no interactive picker, no TUI, no per-repo status.
- A separate \`gwm-workspace\` binary: more surface to maintain; the workspace dimension is orthogonal and fits as a flag.

## Surface

cli (new / changed subcommand)
tui (new view, keybinding, behaviour)

## Backward compatibility

No breaking change — single-repo mode stays the default everywhere.

## Example usage

\`\`\`bash
cd ~/Projects
gwm                              # auto-prompt to open as workspace, or:
gwm --workspace ~/Projects
\`\`\`

In the TUI:

\`\`\`
┌─ workspace ~/Projects (3 repos, 7 worktrees) ─────────────────┐
│   REPO            NAME              BRANCH           STATUS    │
│ ▶ gwm-cli         feat-10-...       feat/#10-...     dirty     │
│   gwm-cli         main              main             clean     │
│   fiches-front    feat-42-search    feat/#42-...     ↑2        │
[...]

PR #609, opened 2026-08-25
✨ feat(config,multiplexer,tui): a setting for whether a mux spawn is a pane, a tab, or a workspace

### LINKED ISSUE #608: [Feature]: a setting for whether a mux spawn is a pane, a tab, or a workspace
## Problem

[#589](https://github.com/kbrdn1/gwm-cli/issues/589) gave `t` a direction, but it folded a second, unrelated choice into the same key. `[tui] mux_pane_direction` currently reads `"right" | "down" | "window"`, where the first two are directions and the third is not: it says *what to open*, not *which half it takes*. The key name is already wrong for one of its three values.

And the list is missing herdr's third structural level. herdr has workspace > tab > pane, and gwm can only reach the bottom two. `herdr workspace create --cwd <path> --label <text> --focus` exists (measured on 0.8.2, same shape as `tab create` minus `--workspace`), so opening a worktree in a workspace of its own is one more arm, not a new mechanism.

## Proposed solution

Split the one key into the two orthogonal choices it is actually carrying:

```toml
[tui]
mux_open_in         = "pane"   # "pane" | "tab" | "workspace"
mux_pane_direction  = "right"  # "right" | "down", only meaningful for a pane
```

- `mux_open_in` replaces the `"window"` value with `"tab"` and adds `"workspace"`. Default `"pane"`, which is what `t` does today.
- `mux_pane_direction` shrinks back to being a direction. Two values, default `right`, still shared with `gwm tmux|zellij|herdr --split`.
- `"tab"` is one thing under three names: a tmux window, a zellij tab, a herdr tab. `Multiplexer::window_noun()` already renders the right word per backend.
- `"workspace"` is **herdr only**. tmux and zellij have no such level, so they **refuse** with a status-bar message rather than opening something else: a silent downgrade to a tab would be the config lying about what it did.

`SpawnMode` grows a `Workspace` variant, and the refusal is a table (which backend can open which target) next to the `macro_refusal` one #589 added.

## Alternatives considered

- **Add `"workspace"` as a fourth value of `mux_pane_direction`.** Smallest diff, but the key name would then be wrong for two of its four values. #589 is unreleased, so renaming costs nothing today and costs a deprecation cycle tomorrow.
- **Fall back to a tab on tmux / zellij.** Consistent with how a `mux_pane` macro degrades to the PTY overlay, but that fallback exists because the macro still has to run *somewhere*. Here there is nothing to preserve: refusing is the honest answer.
- **Give tmux a real session** (herdr workspace ~ tmux session ~ zellij session). It is the exact structural analogue, but it needs two commands (`new-session -d` + `switch-client`) where `spawn_multiplexer` sends one argv, and zellij refuses to nest sessions, so it would be one backend supported out of two.
- **A CLI flag.** `--workspace` is already taken by the global `--workspace <DIR>` ([#36](https://github.com/kbrdn1/gwm-cli/issues/36)), and the CLI already spells the target: bare is a tab, `--split` is a pane. Left out on purpose; a separate ask if the workspace target turns out to be wanted per invocation.

## Surface

config (.gwm.toml schema) + tui

## Backward compatibility

No breaking change **provided it lands before the next release**. `mux_pane_direction = "window"` only exists on `dev` (unreleased, [#606](https://github.com/kbrdn1/gwm-cli/pull/606)), so the value moves to `mux_open_in = "tab"` without any user ever having written it. If a release cuts in between, this becomes a config-schema break instead.

## Example usage

```bash
# .gwm.toml
[tui]
mux_open_in = "workspace"   # `t` opens the worktree in a herdr workspace of its own
```

## Additional context

- Depends on #589 / PR #606; this branch is stacked on it.
- `herdr workspace create --help` on 0.8.2: `--cwd <PATH>`, `--label <TEXT>`, `--env <KEY=VALUE>`, `--focus` / `--no-focus`. No `--workspace` (it is the workspace), so no parent id to pass.
- A `[tui.macro*]` with `open_in = "mux_pane"` refuses under `"workspace"` on every backend: tmux and zellij have none, and herdr's workspace verb takes no trailing command any more than its tab verb does.

kbrdn1:
## Description

**Stacked on #606** (base is `feat/#589-mux-pane-direction`, not `dev`). GitHub retargets it to `dev` when #606 merges; review the two commits here, not the combined diff.

#589 shipped one key carrying two questions. `right` and `down` say which half a pane takes; `window` said *what to open*, which is not a direction. Splitting them makes room for the level gwm could not reach at all: herdr's workspace, above a tab.

```toml
[tui]
mux_open_in        = "pane"    # "pane" | "tab" | "workspace"
mux_pane_direction = "right"   # "right" | "down", pane only
```

Closes #608

## Type of change

## Changes

- **`mux_open_in`** (`pane` / `tab` / `workspace`, default `pane`) picks the level the TUI's `t` key opens. `tab` is the old `window` value under the spelling two of three backends use; `Multiplexer::window_noun` already renders the local word, so a tmux user still reads "window" in the status bar.
- **`workspace` is herdr only, and refused elsewhere.** `herdr workspace create --label <name> --cwd <path> --focus` is `tab create` minus the `--workspace` it would be creating (measured on 0.8.2). tmux and zellij get `tmux has no workspace level: herdr is the only backend with one` on the status bar and open nothing.
- **The refusal lives in the type**: the three builders return `Result<Vec<String>, &'static str>`, so a caller cannot forget a gate, and the sentence names both the backend that cannot and the one that can.
- **`macro_refusal` gains a workspace arm** worded for the level it refuses. A `mux_pane` macro under `workspace` falls back to the PTY overlay on all three backends; the macro path asks `macro_refusal` first so its status names the macro's problem rather than the target's.
- **`SplitDirection` doubles as the config type**, which deletes the parallel `MuxPaneDirection` enum and the `split_direction()` shim that existed only to answer for `window`.
- Docs EN + FR (config reference, multiplexer page, keybindings table), `examples/gwm.toml.example`, and the changelog entry folded with #589's.

## Tests

- [x] `cargo test` passes locally (exit 0, 104 `test result: ok`)

New coverage: the `workspace create` argv (label, cwd, focus, **no** `--workspace` and no `--direction`, and a workspace id making no difference), both refusals naming their backend and herdr, the refusal being per-target rather than per-backend (tmux still opens the other two), `build_command` forwarding the refusal instead of substituting an argv, the macro refusal naming the level, `mux_open_in` defaulting to `pane` and round-tripping its three values, `session` rejected at load, **`mux_pane_direction = "window"` now rejected** so a config written against the unreleased #589 shape fails loudly, the Settings-panel choice lists against both `ALL`s and the new field order, and `t` resolving the two keys to a mode, an argv and a status noun across all three targets.

Checked red: making tmux's `Workspace` arm fall through to the window argv (the silent downgrade this PR exists to avoid) fails `only_herdr_has_a_workspace_level_and_the_others_say_so` with the argv it actually got.

## Checklist

- [ ] README updated if CLI surface changed (the CLI surface is unchanged; `docs/3.cli/` still updated for the herdr verb)

## Linked issues / docs

- Issue: #608
- Related PR: #606 (base; this renames the key it introduces)

## Notes for reviewers

**This must land in the same unreleased cycle as #606, or it stops being free.** `mux_pane_direction = "window"` exists only on `dev`. Renaming it now costs nothing because nobody has written it; renaming it after a release is a config-schema break under the 1.0 promise. If #606 ships without this, the follow-up needs a deprecation path instead.

**Why refuse rather than fall back.** The `mux_pane` macro degrades to the PTY overlay because the macro still has to run *somewhere*. Here there is nothing to preserve: a tab is not a smaller workspace, it is a different thing, and a setting that quietly did something else is the failure mode worth designing against. Both alternatives (silent fallback, and giving tmux a real session) are argued in the issue.

**No CLI flag on purpose.** `--workspace` is taken by the global `--workspace <DIR>` (#36), and the CLI already spells its target: bare is a tab, `--split` is a pane. `mux_open_in` is a `[tui]` key that does not reach the CLI, which is stated in its doc comment and in the reference. A per-invocation workspace flag is a separate ask if it turns out to be wanted.

**herdr is measured, the other two are not, for this target.** `herdr workspace create --help` on 0.8.2 gives `--cwd <PATH>`, `--label <TEXT>`, `--env <KEY=VALUE>`, `--focus` / `--no-focus`. tmux and zellij need no measurement here: the assertion about them is that they have *no* such verb.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,13 +12,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- **A setting for which direction a mux pane opens in**
-  ([#589](https://github.com/kbrdn1/gwm-cli/issues/589)). `[tui]
-  mux_pane_direction` decides where the TUI's `t` key puts the worktree it
-  opens: `"right"` (default) or `"down"` to split the current pane, `"window"`
-  for a whole tmux window or zellij / herdr tab. It is also the direction a
-  bare `gwm tmux|zellij|herdr <pattern> --split` takes, and the new
-  `--direction right|down` overrides it for one invocation. The knob cycles
+- **Two settings for what a mux spawn opens, and where**
+  ([#589](https://github.com/kbrdn1/gwm-cli/issues/589),
+  [#608](https://github.com/kbrdn1/gwm-cli/issues/608)). The TUI's `t` key
+  took whatever each backend felt like giving it. Two `[tui]` keys now say:
+
+  ```toml
+  [tui]
+  mux_open_in        = "pane"    # "pane" | "tab" | "workspace"
+  mux_pane_direction = "right"   # "right" | "down", pane only
+  ```
+
+  `"tab"` is a whole screen of its own: a tmux window, a zellij or herdr tab,
+  one thing under three names. `"workspace"` is herdr's level above a tab and
+  runs `herdr workspace create --label <name> --cwd <path> --focus`.
+  `mux_pane_direction` is also the direction a bare
+  `gwm tmux|zellij|herdr <pattern> --split` takes, and the new
+  `--direction right|down` overrides it for one invocation. Both keys cycle
   live in the Settings panel under the **TUI** tab.
 
   Two directions rather than four: `left` and `up` reach tmux only through
@@ -26,10 +36,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   values: right, down]`, so a fuller compass would be values one backend could
   not honour.
 
-  One caveat, and it is the same shape as the herdr one below: under
-  `"window"` a `[tui.macro*]` with `open_in = "mux_pane"` falls back to the
-  PTY overlay on zellij, because `zellij action new-tab` takes no trailing
-  command to run. The status bar names the backend that refused.
+  **`"workspace"` is refused on tmux and zellij, not downgraded to a tab.**
+  Neither has a level there, and quietly opening something else would leave
+  the setting describing what did not happen. The status bar names the backend
+  that cannot and the one that can. (Both have *sessions*, the structural
+  analogue, but gwm runs inside one: tmux would need two commands to create
+  and switch to a sibling, and zellij refuses to nest sessions.)
+
+  One caveat, the same shape as the herdr one below: a `[tui.macro*]` with
+  `open_in = "mux_pane"` falls back to the PTY overlay under `"tab"` on
+  zellij and under `"workspace"` on every backend, because those verbs take
+  no trailing command to run. The status bar names which one refused.
 
 - **herdr is a third multiplexer backend**
   ([#588](https://github.com/kbrdn1/gwm-cli/issues/588)). `gwm herdr <pattern>`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,13 +14,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Two settings for what a mux spawn opens, and where**
   ([#589](https://github.com/kbrdn1/gwm-cli/issues/589),
-  [#608](https://github.com/kbrdn1/gwm-cli/issues/608)). The TUI's `t` key
+  [#608](https://github.com/kbrdn1/gwm-cli/issues/608),
+  [#611](https://github.com/kbrdn1/gwm-cli/issues/611)). The TUI's `t` key
   took whatever each backend felt like giving it. Two `[tui]` keys now say:
 
   ```toml
   [tui]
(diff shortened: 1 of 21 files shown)


kbrdn1:
Folded into #606 rather than merged on its own: the stacked base was costing more than it bought. `ci.yml` only triggers on `pull_request: branches: [main, dev]`, so this PR never had CI attached and each push needed a manual `workflow_dispatch`; the eventual retarget to `dev` fires an `edited` event, which is not in the default trigger set, so it would likely not have started one either. One PR on `dev` also means #610 rebases once instead of twice.

`feat/#589-mux-pane-direction` fast-forwarded to this branch's head (`e4d07d1`), so nothing was rewritten and every commit is preserved: #589's four, #608's two, the macro-fallback test from the review, and #611.

Closing in favour of #606.

Issue #599, opened 2026-08-25
[Feature]: run a mux_pane macro inside a herdr pane


kbrdn1:
## Problem

`#588` added herdr as a third multiplexer backend, but one of the three call sites deliberately stayed on its old path: a `[tui.macro1]` / `[tui.macro2]` with `open_in = "mux_pane"` still falls back to the PTY overlay under herdr, and now says so in the status bar.

The reason is a shape mismatch, not an oversight. `tui/mod.rs::run_macro` builds one argv and spawns it, appending the macro command to the mux argv (as a single trailing operand for tmux, wrapped in `-- <shell> -c <cmd>` for zellij). herdr has no equivalent:

- `herdr pane split [OPTIONS] [PANE_ID]` takes no trailing command. Its only positional is the pane to split, so appending the command string there would split whichever pane that string resolved to.
- Running a command takes a second call: `herdr pane run <PANE_ID> <COMMAND>...`, and `PANE_ID` only comes back in the JSON `pane split` prints (`.result.pane.pane_id`).

So the herdr arm in the cascade returns `None` with an explicit status rather than building a command that would open an empty pane and drop the macro silently.

## Proposed solution

Give the macro path a two-step form for herdr:

1. spawn `herdr pane split --current --direction right --cwd <path> --no-focus` with captured stdout (the `--no-focus` matters here: the macro's pane is a side effect, the user stays where they are),
2. read `.result.pane.pane_id` off the JSON,
3. spawn `herdr pane run <pane-id> <shell> -c <cmd>`.

This does not fit `multiplexer::build_*_command`, whose contract is a pure `Vec<String>` with no process in it, and that contract is worth keeping: it is what lets `tests/multiplexer_tests.rs` pin every incantation without a multiplexer on the runner. The two-step wants its own seam, e.g. a function that takes the split output as a `&str` and returns either the pane id or a parse error, so the parsing half stays testable against a captured fixture and only the spawn half is untested.

Worth deciding at the same time: what the status bar says when the split succeeds and the run fails, since a bare pane is then already open.

## Surface

tui

## Backward compatibility

No breaking change. Today the case falls back to the PTY overlay, which works; this replaces a fallback with the behaviour the config asked for.

## Additional context

- The herdr arm and its comment are in `src/tui/app.rs` / `src/tui/mod.rs` as of #588 (PR #598).
[...]

Issue #588, opened 2026-08-25
[Feature]: add herdr as a third multiplexer backend


kbrdn1:
## Problem

`src/multiplexer.rs` knows two multiplexers: `Multiplexer::{Tmux, Zellij}`, detected from `$TMUX` / `$ZELLIJ` (`detect_tmux` / `detect_zellij`, lines 96-110). Everything downstream is a two-armed match on that enum:

- `cli.rs:3951-3974` (`cmd_multiplexer`, behind `gwm tmux` / `gwm zellij`),
- `tui/app.rs:5154-5170` (`open_in_mux_pane`, the `t` key),
- `tui/mod.rs:1414-1423` (a `open_in = "mux_pane"` macro, with the PTY overlay as fallback).

[Herdr](https://herdr.dev) is a third terminal multiplexer, built around coding agents, and gwm already models agent sessions (#408). Running gwm inside Herdr today, `t` reports `no multiplexer detected ($TMUX / $ZELLIJ not set)` and the macro path falls back to the PTY overlay.

## Proposed solution

Add `Multiplexer::Herdr` with the same three pieces the other two have: a detector, an argv builder, and the arms in the three call sites, plus a `gwm herdr <pattern>` subcommand for parity.

Measured against the installed CLI (`herdr 2026-08`, `herdr --help`, `herdr pane`, `herdr tab`, `herdr worktree`):

- **detection**: `$HERDR_ENV` is `1` inside a managed pane, alongside `HERDR_PANE_ID`, `HERDR_TAB_ID`, `HERDR_SOCKET_PATH` and `HERDR_BIN_PATH`. `$HERDR_ENV` is the direct analogue of `$TMUX`.
- **`SpawnMode::Window`**: `herdr tab create --cwd <path> --label <name> [--focus]`.
- **`SpawnMode::Split`**: `herdr pane split --current --direction right|down [--ratio FLOAT] --cwd <path> [--focus]`.
- **worth evaluating instead of the generic pair**: `herdr worktree open --path <path> [--label TEXT] [--focus]` is a first-class worktree primitive, so Herdr can label and track the pane as a worktree rather than as a bare cwd. It takes `--branch NAME` too. If it behaves the way the name suggests, it is the better builder for gwm and the one the issue should ship.

`herdr pane split` requires `--direction`, which tmux and zellij leave implicit. That is the same knob as the pane-direction setting filed separately; whichever lands first should pick the default (`right` reads as the natural analogue of tmux `-h`).

[...]

PR #610, opened 2026-08-25
✨ feat(tui,config): `o` on the agents overlay resumes the session in a mux pane

### LINKED ISSUE #591: [Feature]: `o` on the agents overlay opens the session in a mux pane (multiplexer only)
## Problem

The agents overlay (`a`, `Action::AgentSessions`, #408) lists the AI-agent sessions gwm detected per worktree. Its modal keymap has four verbs (`src/tui/modal_keymap.rs:293-297`):

```
DetailAttach  => "attach"       [ "a" ]
DetailDetach  => "detach"       [ "d" ]
DetailInput   => "attach_by_id" [ "i" ]
DetailClose   => "close"        [ "Esc" ]
```

None of them opens the session. `attach` is a **pin**: `attach_selected_agent` (`tui/app.rs:3543`) writes `branch.<b>.gwm-agent-pin` in git config so gwm attributes that session to that worktree. It changes gwm's bookkeeping, not where the session runs.

So after reading which agent is working where, going there means leaving the overlay, moving the selection to the right row, pressing `t` for a mux pane, and then resuming the session by hand.

`o` is unbound in the agents context. In the list view it is `Action::TerminalPty`, so the letter already reads as "open a terminal here".

## Proposed solution

Bind `o` in the agents context to **open a multiplexer pane running the selected session**, at that session's recorded `cwd`.

Two properties, both deliberate:

- **Multiplexer only.** With no multiplexer detected (`$TMUX` / `$ZELLIJ`, plus `$HERDR_ENV` if #588 lands), `o` does nothing but say so. No PTY-overlay fallback: the point is to put the session next to gwm, and an overlay that covers gwm is not that.
- **The session, not a shell.** The pane resumes the agent conversation, it does not just `cd` there. `AgentSession` (`src/agent_sessions.rs:59-76`) already carries everything needed: `kind`, `cwd`, `id`, `ended`, `name`.

Resume incantations, measured against the installed binaries (2026-08-25), not assumed:

| kind | command |
|:--|:--|
| `ClaudeCode` | `claude -r <session-id>` (`--resume` takes a session id or opens a picker) |
| `Codex` | `codex resume <SESSION_ID>` (uuid or session name) |
| `opencode` | `opencode -s <session-id>` (`--session`) |
| `Vibe` | `vibe --resume <SESSION_ID>` |

These are four third-party CLIs on their own release cadence, so the mapping belongs in config with those as defaults, in the vocabulary `[tui.open] editor_cmd` / `shell_cmd` already use:

```toml
[tui.agent_resume]
claude = "claude -r {session}"
codex  = "codex resume {session}"
```

The pane itself reuses the `mux_pane` builders and the pane-direction setting (#589): same spawn path as `t`, with a command instead of a bare shell.

## Surface

tui (new keybinding in the agents context) + config (.gwm.toml schema)

## Backward compatibility

No breaking change: `o` is currently unbound in that context, and it joins `[tui.keys.modal.detail]` like the other three, so it stays rebindable.

## Additional context

- `open_in_mux_pane` (`tui/app.rs:5153`) reads `self.selected()`; here the target is the overlay row, and `detail_overlay_target` already exists for the agents consumer (`tui/app.rs:3189` clears it on the CI-checks path).
- A session with `ended = true` still resumes, that is what resume is for. A **live** session is the interesting case: resuming it in a second pane while it runs elsewhere may fork or refuse depending on the tool. Worth measuring per agent before shipping, and worth a status message rather than a silent second pane.
- The `#453` help-overlay completeness guard picks up a new `ModalAction` automatically, so the overlay documents `o` without a second edit.
- gwm already sanitises what it hands to a shell (see GHSA-fffq-vg6f-gxqm, branch names in hooks). A session id and a session name come from a third-party artefact on disk, so they go through the same neutralisation before reaching the pane command.

kbrdn1:
## Description

The agents overlay tells you which agent is working where and then leaves you to get there by hand: close it, move the selection to the right row, press `t`, resume the conversation yourself. `a` did not help, because it is a **pin**: it changes gwm's bookkeeping, not where the session runs.

`o` now opens a multiplexer pane running the selected session.

Rebased onto `dev` after #606 landed, so this sits on the post-#589 / #608 multiplexer API (`detect_multiplexer` / `macro_refusal` / `build_command`) rather than the `detect_split_command` it was first written against.

Closes #591

## Type of change

- [x] ♻️ Refactor (no behaviour change) — `attach_pane_command` extracted, see below
- [x] 📝 Docs
- [x] ✅ Tests

## Changes

- **`attach_pane_command` in `src/multiplexer.rs`.** #589 took the first half of `run_macro`'s inline chain out as `macro_refusal` ("can this (backend, mode) run a command"); the second half — the argv assembly that knows zellij runs its trailing argv directly while tmux takes one shell-command operand — stayed behind because it had a single caller. `o` is the second, so the two halves now sit together.
- **`[tui.agent_resume]`** with four defaults: `claude -r {session}`, `codex resume {session}`, `opencode -s {session}`, `vibe --resume {session}`. Configuration rather than a hardcoded table because these are four third-party CLIs on their own release cadence.
- **`expand_agent_resume`** — single-pass, shell-quoting `{session}`. A session id is read out of a third-party artefact on disk, so it is data; same rule the hook placeholders learned in GHSA-fffq-vg6f-gxqm.
- **`ModalAction::DetailOpenPane`** (`o`, verb `open_pane`) under `[tui.keys.modal.detail]`, so it is rebindable like the other three. Help overlay row and hint-bar entry added.
- **`plan_agent_pane`** — everything decided before a process is spawned, split out so it is testable (same reason `mux_pane_status` was).
- `platform_shell` extracted from `run_macro`; `agent_pane_status` builds on `mux_pane_status` including its new `noun`, so `mux_open_in = "tab"` does not leave the status describing a pane.

## Three things worth reading closely

**The pane opens in the overlay's worktree, not at `session.cwd`.** This deviates from the issue's wording, deliberately. The overlay lists **pinned** sessions too, and a pin exists exactly when the recorded directory names the wrong tree, which is the documented reason `gwm agents attach` exists. `overlay_pins` leaves `cwd` alone "purely as provenance", so for a pinned Claude session resolved by the id sweep it is not a worktree at all: it is the slug directory under `~/.claude/projects`. Resuming there drops the agent inside its own artefact store. An auto-matched row has `cwd == target` anyway, so the overlay target is right in both cases and the pinned case is only right this way. Backing the fix out makes the test report `tmux split-window -h -c /home/u/.claude/projects/-home-u-main-checkout`.

**`o` reads `mux_open_in` / `mux_pane_direction`, so it has three refusal families, not one.** The issue predates #608 and only promised the direction setting, but two keys opening at different levels from the same overlay would be the odd outcome. That means the refusal routes through `macro_refusal` rather than a hardcoded herdr string: a zellij **tab** takes no command either, and so does neither herdr level.

**No PTY-overlay fallback.** The point of `o` is to put the session *next to* gwm; an overlay that covers gwm is not that. So outside a multiplexer the key says so and does nothing, with the same sentence `t` shows (naming all three env vars) rather than the macro path's shorter "no multiplexer".

## Tests

- [x] `cargo test` passes locally (104 suites green)
- [x] Each of the four commits verified green individually, so a bisect does not land on a red tree

No test here spawns a process: every one refuses before the spawn, so none needs tmux, zellij or herdr on the runner. Three guards were each verified to go **red** with their fix backed out, not merely observed green — the recorded-cwd one, the zellij-tab refusal (which otherwise reaches the spawn), and herdr's (which otherwise reports "opened codex session in new pane" for a pane running nothing).

The `macro_refusal` / `attach_pane_command` agreement test is an **implication**, not an equivalence, and I wrote it as an equivalence first: zellij tabs refuse in the former and build in the latter, on purpose, because teaching the mode to `attach_pane_command` would be the second copy the extraction exists to avoid.

## Screenshots / TUI captures

None. The observable half is a new pane in the surrounding multiplexer, which a `script(1)` capture of gwm cannot show. **This needs a local install check before merge** — I have not driven the key against a real tmux pane.

## Checklist

- [ ] README updated if CLI surface changed — no CLI surface change, TUI + config only

## Notes for reviewers

- **The four resume incantations were re-measured** against the installed binaries rather than taken from the issue body: `claude -r, --resume [value]`, `codex resume [SESSION_ID]` (UUID or session name), `opencode -s, --session`, `vibe --resume [SESSION_ID]`.
- **Not done, on purpose: the Settings panel does not expose these four fields.** #589 did add its two settings there, so the precedent cuts both ways; the difference is that those cycle through a closed enum while these are free-text command lines, and `[git_tui] command` / `[review] command` are not in the panel either. There is no completeness guard over `SettingField`. Say so if you disagree and I will add them.
- **Not done: what each CLI actually does when you resume a session that is still live.** The issue asked for that measurement; it needs four running sessions and a second pane each, and the answers would still be per-version. Instead a live row is flagged on the status bar ("still active elsewhere, the agent may fork or refuse"), which is the fallback the issue itself named. A session with `ended = true` resumes silently.
- The macro path lost its inline argv assembly to the shared builder. Behaviour unchanged, including the refusal status and the PTY fallback; a backend answering `None` lands in that fallback rather than panicking.

## Linked issues / docs

- Issue: #591
- Related: #588 (herdr backend), #589 / #608 (the mux level and direction settings this now reads, merged in #606), #599 (the herdr pane-id round trip that would let herdr run a command)


kbrdn1:
Heads-up from #606 / #609: this branch will stop compiling the moment #606 merges, and the two of us are duplicating one concern. Both are mechanical to resolve, but neither shows up as a merge conflict, so flagging before either lands.

**`detect_split_command` is deleted by #606.** It splits into `detect_multiplexer(tmux, zellij, herdr) -> Option<Multiplexer>` and `build_command(mux, name, path, mode, workspace) -> Result<Vec<String>, &'static str>`, because a third call site (the CLI verb) had the backend match written out a second time. This branch uses the old function in six places:

- `src/tui/app.rs:563` (in the diff) — `let Some((mux, split)) = crate::multiplexer::detect_split_command(kind, target, tmux, zellij, herdr)`
- `tests/multiplexer_tests.rs` — the import, plus three call sites in `attach_pane_command_*`
- three doc-comment references in `attach_pane_command`'s rustdoc

The rewrite at each site:

```rust
// before
let Some((mux, split)) = detect_split_command(kind, target, tmux, zellij, herdr) else { ... };

// after
let Some(mux) = detect_multiplexer(tmux, zellij, herdr) else { ... };
let split = build_command(mux, kind, target, mode, workspace)?;   // mode from the config, see below
```

**`attach_pane_command` and `macro_refusal` answer the same question.** #606 added:

```rust
pub const fn macro_refusal(mux: Multiplexer, mode: SpawnMode) -> Option<&'static str>
```

for exactly the case this PR's `Multiplexer::Herdr => return None` covers: which (backend, mode) pairs have no trailing-command form. It also covers two your version cannot see, because they did not exist yet: `zellij action new-tab` under `mux_open_in = "tab"`, and every backend under `mux_open_in = "workspace"` (#609). Keeping both means the herdr answer is written twice and the other two are missing from one of them, which is the defect #588 extracted the cascade to kill in the first place.

Suggested shape: `attach_pane_command` keeps the argv *attachment* (the tmux-single-operand vs zellij-`-- shell -c` split, which is genuinely its own knowledge) and drops the herdr arm, with the caller asking `macro_refusal` first. That is what `run_macro` does on #606, so the two paths end up identical.

**`SpawnMode` also grew.** It is `Window | Split(SplitDirection) | Workspace` now, so a `Split` argv is no longer the only thing a call site can be handed, and `[tui] mux_open_in` decides which one `t` builds. An agent-resume pane presumably always wants a `Split`, in which case it should say so explicitly rather than inherit the key.

**Order.** #606 is the oldest of the three, reviewed clean with 13 green checks; #609 is stacked on it. Merging #606 → #609 → rebasing this one is the cheapest path, and the rebase is the two edits above. Happy to do the rebase here instead if you would rather this one goes first, but then #606 and #609 pay the same cost twice over.

Per CONTRIBUTING, this is the "parallel agents only when file ownership is disjoint" case: both branches touch `src/multiplexer.rs`, `src/config.rs`, `src/tui/app.rs`, `src/tui/mod.rs`, the same three test files, `docs/4.configuration/1.gwm-toml.md` (+ fr), `examples/gwm.toml.example` and `CHANGELOG.md`.

kbrdn1:
@codex review

kbrdn1:
Rebased onto `dev` after #606, and both Copilot findings applied. Answering the heads-up above point by point, including the two places I went the other way.

## The heads-up from #606 / #609

All three mechanical points are done: `detect_split_command` is gone from this branch (`detect_multiplexer` + `build_command` at every site, imports and rustdoc included), and `attach_pane_command` no longer answers "can this backend run a command" — `macro_refusal` does, so the herdr sentence is written once and the two cases it could not see (zellij tab, workspace) now refuse correctly. There is a test for the zellij-tab one, which this branch had no way to produce before the rebase.

**Where I went the other way, twice.**

1. **`SpawnMode`: `o` inherits `mux_open_in` rather than pinning `Split`.** The suggestion was that a resume pane "presumably always wants a Split". The counter-argument that decided it: `t` and `o` sit on the same row, and having one open a tab while the other opens a pane is the kind of inconsistency a setting is supposed to remove, not create. The cost is real and accepted — under `mux_open_in = "workspace"` a tmux user loses `o` entirely, and the status bar has to say so. That is the same thing `t` already does there.

2. **`attach_pane_command` keeps its herdr arm returning `None`.** The suggestion was to drop it now that `macro_refusal` gates the call. Kept because appending an operand `herdr pane split` ignores opens an empty pane and drops the command *silently*, and that failure is worth being unrepresentable rather than merely unreachable. The agreement test pins the implication that actually matters: nothing `macro_refusal` lets through can fail to build. The converse is deliberately not owed — a zellij tab refuses in the first and builds in the second.

Both are cheap to reverse if you disagree; say which and I will.

## Copilot findings

**`shell_words::quote` is not enough — confirmed, fixed.** The finding is correct and the reasoning is worth keeping: quoting only holds for an *unquoted* POSIX position, and the template around the placeholder is the user's text. `claude -r "{session}"` is a natural override, and there the single quotes are literal:

```
$ sh -c 'echo "'"'"'$(echo PWNED)'"'"'"'
PWNED
```

`cmd.exe` never honours them at all, so on Windows the quoting was decorative. Since where the placeholder sits is not ours to decide, the value is constrained instead: `is_shell_safe_session_id` admits ASCII alphanumerics plus `-`, `_`, `.`, and `expand_agent_resume` returns `None` otherwise, surfacing as a refusal rather than a line nobody can predict. Every id the four backends produce is a UUID or a slug, so nothing legitimate is refused. Quoting stays as the second layer. Structured argv was the other option and was not taken: `[tui.agent_resume]` is a shell line by the same contract as `[git_tui] command` and `[tui.macro*]`, and changing that shape for this one table would be worse than constraining the one value it interpolates.

**The success path was untested — correct, fixed.** Every test stopped at a refusal or at pure argv planning, so a regression that plans correctly and never launches would have passed. There is now a spawn test against a recording fake `tmux` on `PATH` (the `cli_binary.rs` shape, `#[cfg(unix)]`, under this binary's `env_lock()`), asserting the argv reaches the multiplexer, carries the resume command rather than a bare shell, and that the status is built from the session's own freshness. Backing the spawn out turns it red with an empty log.

## Still open

Codex review has not run on this branch — both the cloud bot and the local CLI are on the same exhausted usage quota. This has only been through Copilot.

And it still wants a local install check: I have not driven `o` against a real tmux pane, and a capture of gwm cannot show the pane next to it.

Issue #594, opened 2026-08-25
[Feature]: modals should follow `[tui] layout` instead of always being bordered


kbrdn1:
## Problem

`[tui] layout = "compact"` has been the default since #545: one filled header line per section, no rules, which saves one row and one column of chrome per pane against `bordered`. The list view, the sidebar and the panes all honour it (`app.config.tui.layout.is_compact()` at `tui/ui.rs:337`, `377`, `937`, `1004`).

Modals do not. `overlay_block` (`tui/ui.rs:5852-5858`) is unconditional:

```rust
Block::default()
  .borders(Borders::ALL)
  .border_type(BorderType::Rounded)
  .padding(Padding::symmetric(2, 1))
```

so every modal costs 2 rows and 2 columns of border plus 2 rows and 4 columns of padding, whatever the layout says. That was a deliberate call at the time ("a modal keeps its rules whatever `[tui] layout` says: a panel floating over content is exactly where a border earns its keep", `draw_report`), and this issue is the request to revisit it.

## Proposed solution

Make the modal frame follow `[tui] layout`:

- `bordered`: today's rounded box with symmetric padding.
- `compact`: the filled header line the panes already use, no side rules, tighter padding.

The whole thing funnels through `overlay_block` / `overlay_block_titled`, so the frame itself is one function. What needs care is every caller that sizes itself off the frame: the doc comment on `overlay_block_titled` states the sizing contract callers follow (`lines.len() + 2 /* border */ + 2 /* padding */`), and #550 already showed what happens when a modal's width rule changes under a consumer that recomputed it by hand.

## Surface

tui (rendering)

## Backward compatibility

Behaviour change for anyone on the default layout: modals lose their box. It is opt-out through `layout = "bordered"`, which is the same escape hatch #545 shipped.

## Additional context

- The regression surface is the modal render test suite (`tests/tui_modal_render_tests.rs`), which pins exact widths per modal in a characterisation matrix (`sizing_matrix`). Those numbers are the contract: expect to add a compact column rather than to change the bordered one.
- `modal_rect` in those tests locates the frame by its border glyphs. A compact modal has no `╰` to find, so the helper needs a second locator before the suite can assert anything about it (see #553 for what a mis-located `modal_rect` does to a red test).

kbrdn1:
Landed on `dev` in #616 (merge `cb5126d`). `dev` is not the default branch, so the `Closes` keyword does not fire on merge; closing by hand.

Issue #593, opened 2026-08-25
[Feature]: a keybinding that opens the commit listing full size, with load-more


kbrdn1:
## Problem

The Recent Commits pane shares the sidebar with the identity, status and working-tree blocks, and it renders a **bounded** list: `recent_commits_lines(w, limit, theme)` (`tui/ui.rs:2119`) asks `worktree::recent_commits_cached(w, limit)` (`src/worktree.rs:1139`) for a fixed number of rows. There is no way to see more history without leaving gwm for lazygit (`l` / `L`).

The commit graph renderer (`src/tui/commit_graph.rs`) already draws the `○` / `◎` nodes and the connectors, so the missing piece is a surface big enough to use it and a way to ask for more.

## Proposed solution

A keybinding that opens the commit listing as a full-size modal:

- new `Action::CommitsModal` with a default chord (see the note on the numeric family in the Working Tree modal issue),
- the same graph renderer over the full area,
- **load more**: a key (or reaching the bottom) raises the limit and re-reads, so history is paged rather than capped. `recent_commits_cached` is keyed by `(repo, tip, limit)`, so a larger limit is a fresh entry rather than an invalidation.
- `Enter` on a row is a natural follow-up (copy the sha, open the commit on the forge), but it is not required for the first cut.

## Surface

tui (new view + keybinding)

## Backward compatibility

No breaking change: additive action and modal, sidebar pane untouched.

## Additional context

- Paging changes the cache's shape: today the limit is effectively constant, so the map holds one entry per worktree tip. Paging multiplies entries per worktree. Worth capping the page count or evicting the smaller limits.
- The revwalk is `recent_commits_revwalk` (`src/worktree.rs:1182`); a deep page on a large repo is exactly the kind of work the async-task spine (#231) exists for, so the load-more should go off-thread rather than block the frame.
- Ship this with the Working Tree modal if one shell can serve both.

kbrdn1:
Delivered in #614, merged into `dev`.

Closing by hand: the PR carries the `Closes` keyword, but GitHub only links it for pull requests targeting the default branch, and this one targeted `dev`. Ships with the next release.

Issue #343, opened 2026-06-26
[Task]: Move synchronous git subprocesses off the TUI render path


kbrdn1:
## Scope

Two `git` subprocesses run **synchronously inside the TUI render path**, stalling navigation on large repos / slow filesystems:

- [ ] `git_diff_stat_vs_base` (`src/tui/ui.rs:658`) — runs inside `draw_sidebar`, guarded by a **single-slot** cache, so each selection change (`j`/`k`) is a cache miss that shells out **inside `terminal.draw()`** (A→B→A re-runs). Already moved off per-frame (comment `:656-657`) but not off-thread.
- [ ] `git_status_short` (`src/tui/ui.rs:1135`, `git status --porcelain -z`) — same render surface; one caller (`:950`), per-selection.
- [ ] `maybe_auto_refresh` (`src/tui/app.rs:1046`) calls `refresh()` **synchronously in workspace mode**, while the single-repo path is off-thread via `spawn_refresh` (`:1067`).

Route all three through the existing `TaskRunner` (#231): render the last-known value, refresh async (keyed by path), debounce `j`/`k`.

## Motivation

The only real UX-regression risk in an otherwise mature TUI — the last unfinished step of the off-thread migration that #231 started.

## Kind

perf (measured improvement)

## Acceptance criteria

- [ ] No git subprocess runs inside `terminal.draw()`
- [ ] Sidebar diff/status and workspace auto-refresh ride the `TaskRunner`
- [ ] Navigation stays responsive on a large repo (manual check + a state-machine test where feasible)

## Notes / links

From the v1.0.0 ultra-audit (quality dimension). Per project memory, the per-frame sidebar clone was already removed — this is the remaining synchronous-subprocess tail.

kbrdn1:
Livré par #351 (mergée dans `dev`, merge commit 5790d8f). Les deux subprocess git synchrones restants (sidebar + refresh workspace) sont sortis du render path via le TaskRunner (#231) : `TaskKind::Sidebar` + `TaskKind::RefreshWorkspace`, le render ne shell plus (placeholder `loading…` sur cache stale), coalescing = debounce. TDD anchor red-first + state-machine déterministes. Review Codex clean (2 itér), CI verte 3 OS.

Issue #219, opened 2026-06-06
[Feature]: make modal TUI keys rebindable with contextual actions


kbrdn1:
## Problem

PR #218 improves the TUI modal ergonomics, but several modal-local keys remain hard-coded in `src/tui/mod.rs` instead of going through `[tui.keys]`.

Examples:

- create modal: `Tab` / `Shift+Tab`, `Enter`, `Esc`, type cycling, text input guards
- confirm modal: `y` / `n`, `h` / `l`, arrows, `Tab`, `Enter`, `Esc`
- link prompt: `i` / `p`, `j` / `k`, arrows, `Enter`, `Esc`, digit input
- other overlays: help/report/open-menu/palette close and activation keys

Today the public keymap intentionally leaves contextual keys out. `src/tui/keymap.rs` documents `Esc` / `Enter` as hard-coded escape hatches because their meaning depends on the active view and sometimes the modal stage. That keeps issue #87 simple, but it means the TUI is only partially rebindable.

## Proposed solution

Introduce a contextual TUI action keymap for modal and overlay keys.

Expected shape:

- Add typed `Action` coverage for modal-local verbs, not raw keys: e.g. `modal_cancel`, `modal_submit`, `modal_next`, `modal_prev`, `confirm_yes`, `confirm_no`, `link_issue`, `link_pr`, `link_next`, `link_prev`, etc.
- Add a context grammar so the same physical key can mean different actions in different places without global conflicts.
- Keep the event loop deterministic: no runtime chord timeout and no ambiguity that depends on wall-clock timing.
- Preserve hard emergency exit behavior where needed, especially `Ctrl+C`.
- Update `gwm tui keys`, help overlay rows, and `gwm doctor` so modal bindings are discoverable and validated.
- Keep the implementation TDD-first, with pure state/handler tests before production routing changes.

A possible config direction, to validate before implementation:

```toml
[tui.keys.list]
quit = ["q"]
open_menu = ["o"]

[tui.keys.create]
next_field = ["Tab"]
prev_field = ["BackTab"]
submit = ["Enter"]
cancel = ["Esc"]

[tui.keys.confirm]
confirm = ["y"]
cancel = ["n", "Esc"]
focus_confirm = ["Left", "h"]
focus_cancel = ["Right", "l"]
activate = ["Enter"]

[tui.keys.link.choose_target]
next = ["j", "Down"]
prev = ["k", "Up"]
issue = ["i"]
pr = ["p"]
[...]

PR #612, opened 2026-08-26
✨ feat(tui): the Working Tree listing full size, and the toggle key that closes it

### LINKED ISSUE #592: [Feature]: a keybinding that opens the Working Tree listing full size
## Problem

The Working Tree pane (#300) renders the selected worktree's changed files as a tree inside the sidebar. The sidebar is a fraction of the screen, and the pane is one of several blocks sharing it, so a worktree with more than a handful of changed files scrolls two rows at a time through `J` / `K` (`Action::WtScrollDown` / `WtScrollUp`, `src/tui/keymap.rs:505-507`).

There is no way to see the whole change set at once, on a screen that has the room for it.

## Proposed solution

A keybinding that opens the Working Tree listing as a full-size modal, the way `3` opens the command logs and `4` opens the Settings panel:

- new `Action::WorkingTreeModal` (name to bikeshed) with a default chord,
- a modal at the same size class as the command-log transcript (a full-bleed surface, see `centered` / the "three text canvases" note in `tui/ui.rs`),
- the existing `wt_tree.rs` renderer, given the whole area instead of a sidebar block,
- `j` / `k` (or the arrows) scroll, `Esc` closes, through a `[tui.keys.modal.*]` context like every other modal.

The tree building already lives in `src/tui/wt_tree.rs` and is independent of the pane it draws into, so this is a new view over an existing model rather than new logic.

## Surface

tui (new view + keybinding)

## Backward compatibility

No breaking change: additive action and modal. The sidebar pane stays exactly as it is.

## Additional context

- Free single-key chords are getting scarce in the list view; the numeric family (`1` `2` `3` `4`) is where overlays live, so `5` is the obvious candidate and keeps the "numbers open panels" convention.
- Same question as the commits modal filed alongside this one; the two should look and behave alike, so land them with one shared shell if the shape allows.

### LINKED ISSUE #613: [Bug]: an overlay's toggle key stops closing it once rebound to a chord, or to a key its own context claims
## Description

Every numeric overlay closes on its own opener: `3` shuts the Command
Logs, `4` the Settings panel, `5` the Working Tree listing (#592). The
dispatch does it with one guard arm at the end of the modal match:

```rust
_ if app.key_matches_action(key, Action::CommandLogs) => app.view = View::List,
```

That arm has two holes, and both are silent.

**It runs after the modal resolution.** Rebind `command_logs` (or
`config_panel`, or `working_tree`) to a key the overlay's own context
already claims, and the modal verb wins: the key opens the overlay from
the list, then scrolls inside it instead of closing. `j` is the obvious
example, but `g`, `G`, `q` and the arrows are all claimed by the
scroll-only contexts.

**`key_matches_action` only tests the current single stroke.** It calls
`keymap.lookup(&[KeyStroke::from_event(&key)])`, so a multi-stroke chord
(`working_tree = ["g w"]`) can open the overlay through the normal
chord-aware list dispatch, but can never match here. The overlay opens
and the same chord does nothing.

Reported by Copilot on PR #612. Not introduced there: `git show
origin/dev:src/tui/mod.rs | grep key_matches_action` finds the same
pattern on four arms already (`CommandLogs`, `ConfigPanel`,
`ToggleDeleteBranch`, `FetchGithub`), so #592 added a fifth instance of
an existing shape rather than a new bug. Filed separately because the
fix changes the resolution order for every modal, which is more than
that PR's contract.

## Steps to reproduce

1. In a repo's `.gwm.toml`:

   ```toml
   [tui.keys]
   working_tree = ["j"]
   ```

2. `gwm` → `j` opens the Working Tree overlay.
3. Press `j` again.

And the chord half:

1. `working_tree = ["g w"]`
2. `gwm` → `g w` opens the overlay.
3. Press `g w` again.

## Expected behaviour

The opener closes the overlay, whatever it is bound to. That is what the
docs promise: "`5` closes it too" / "`3` toggles it shut even when
rebound".

## Actual behaviour

Case 1: the second `j` scrolls the listing down one row. The overlay
never closes on its opener; `Esc` / `q` still work.

Case 2: `g w` does nothing at all. `key_matches_action` sees only the
`w`, which resolves to nothing, and the pending-chord buffer is not
consulted inside the modal arms.

## Severity

🟡 Minor (annoying but can be worked around)

## Surface

tui (key dispatch)

## Additional context

Three shapes the fix could take, in rising order of how much they move:

- Resolve the global toggle **before** the modal verbs for the arms that
  advertise a toggle. Cheapest, but it means an opener bound to `j`
  makes `j` unusable for scrolling in that overlay, which is arguably
  the right trade since the user asked for it.
- Route the modal arms through the chord-aware dispatch so a multi-stroke
  opener resolves. Fixes the second half, not the first.
- Refuse the binding at load time, the way
  `KeyContext::reserved_typing_stroke` already refuses a verb bound to a
  stroke an always-typing context eats. Turns a silent hole into a
  message that names the conflict.

Whichever lands should cover all five arms at once, plus a test per
overlay, otherwise the next overlay copies the shape again.

kbrdn1:
## Description

`5` opens the sidebar's Working Tree pane as a full-size overlay. Same
file-explorer tree, same nerd-font icons, same per-category colours, the
same `<glyph> <n>` change counts on the bottom rule. The difference is
that a change set of more than a handful of files reads in one go instead
of two rows at a time through `J` / `K`.

Review then turned up a bug underneath it, filed as #613 and fixed here
at the user's request: the guard that lets an overlay close on its own
opener never worked for a rebound key. That half touches the command logs
and the settings panel too.

Closes #592
Closes #613

## Type of change

- [x] ✨ Feature (new functionality) — the overlay, #592
- [x] 🐛 Fix (bug fix) — the toggle key, #613
- [x] ♻️ Refactor (no behaviour change) — routing lifted into `App` methods

### The overlay (#592)

- `Action::WorkingTree` (`working_tree`, default `5`), in the numeric
  family with `1` `2` `3` `4`, plus its command-palette entry and its
  help-overlay row.
- `View::WorkingTree` + `KeyContext::WorkingTree`, five verbs under
  `[tui.keys.modal.working_tree]`: `close`, `scroll_down`, `scroll_up`,
  `scroll_top`, `scroll_bottom`. Its own context rather than a share of
  `command_logs`, so rebinding one does not silently rebind the other.
- `WorkingTreeModal` state (`src/tui/state/working_tree.rs`): the
  snapshotted rows, their counts, and the scroll cursor with its
  renderer-published bound.
- `draw_working_tree`: the command-logs shell (fixed title, scrollable
  body with scrollbar, fixed footer hints) with the change counts
  right-aligned on the bottom rule, where the bordered sidebar pane puts
  them.

**The one design call worth reviewing.** The rows are snapshotted in
`enter_working_tree` rather than read from `SidebarState::cache`. That
cache is keyed by `(path, mode)` and is only rebuilt while the sidebar is
**open** and in **commits** mode, so reading it would leave the overlay
blank in exactly the two states where seeing the change set is most
useful: sidebar hidden, or the Details panel showing stashes.
`tui_app_tests` closes the sidebar before opening, so a cache-reading
version goes red there.

**The read runs on a worker**, not on the keypress. It shipped inline at
first, argued as bounded because `git_status_short` kills git at
`STATUS_SCAN_CAP` records. That was the wrong reassurance: the cap bounds
how many records git yields, not how long it takes to produce the first
one, so an untracked walk on a cold or network filesystem froze the event
loop for its whole duration. `TaskKind::WorkingTree` +
`TaskMsg::WorkingTree` now carry it, the shape the sidebar preview has
had since #343; the overlay opens on a loader and fills in when the
worker lands.

Coalescing on that slot is per-path: a repeated `5` on the same worktree
rides the read already out, while a reopen on a different one invalidates
it, because its payload would be dropped on the path check and the loader
would have nothing left to clear it.

It is not routed through `build_sidebar_sections`, which would also run
`git log`, `git stash list` and the diff-vs-base stat for panes the
overlay does not show.

### The toggle key (#613)

`3`, `4` and `5` each close the overlay they open, but the guard doing it
asked `key_matches_action` (a single-stroke lookup) and only ran after
the modal verbs. Two silent holes:

- `working_tree = ["g w"]` opened the overlay through the chord-aware
  list dispatch and could never shut it;
- `working_tree = ["j"]` opened it and then scrolled it, because the
  modal verb resolved first.

- `Keymap::chords_for` is the inverse lookup `lookup` cannot do: what is
  bound to *this* action.
- `App::modal_toggle_stroke` walks it against the pending buffer and
  answers `Fired` / `Pending` / `Unclaimed`, so a chord accumulates and
  its prefix is **consumed** rather than handed to a scroll verb (`g`
  would otherwise jump the listing to the top on its way to `g w`).
- Deliberately **not** `dispatch_key`: that resolves the whole keymap,
  and `d` opening the delete confirm from behind an overlay would be
  worse than the bug being fixed.

Applies to all three overlays. The settings panel keeps its capture and
edit sub-modes ahead of the toggle, since those own every stroke while
live. `ToggleDeleteBranch` / `FetchGithub` keep the old guard: they are
global actions reused inside a modal, not toggles, and nothing promises
they close anything.

### Routing

`handle_working_tree_key`, `handle_command_logs_key` (returning
`CommandLogsKey`, since one outcome is a clipboard write the run loop
owns) and `handle_config_nav_key` lift each overlay's dispatch into a
testable `App` method, the shape the create overlay has had since #217.
That is what makes the precedence assertable at all: a `match` in the run
loop cannot go red.

### Docs

Keybindings table + a section per language, the config reference's modal
context list and rebindable-actions table, and a CHANGELOG entry under
`[Unreleased]` in both Added and Fixed.

## Tests

- [x] `cargo test` passes locally (104 test binaries ok, run under a
  CI-like minimal `PATH`)

Twenty-four new pins. The ones worth naming:

- **The cache trap** — the snapshot with the sidebar closed, which is the
  state that would have shipped an empty overlay.
- **Precedence, once per overlay** — with the toggle rebound onto a
  scroll key, the overlay closes *and* the listing does not move. Each of
  the three was verified red by swapping the two blocks back.
- **Chords** — the prefix is consumed without firing `scroll_top`, and a
  stray prefix is dropped rather than left armed to complete a chord the
  user never typed.
- **Blast radius** — `d` / `n` / `x` / `3` / `4` all come back
  `Unclaimed` from the working-tree toggle, which is the invariant that
  rules out `dispatch_key`.
- **Placement, not just presence** — the change counts are scanned on the
  modal's last row with rule to their left and nothing but padding to
  their right. Verified red by dropping `.right_aligned()`.
- **The sizing matrix** entry (a text canvas like the transcript: 90%, no
  ceiling), verified red on a wrong pin rather than left to sit unread.
- **The async slot, enumerated** — `the_overlay_never_waits_on_a_worker_that_is_not_there`
  walks all four ways into the overlay (nothing selected / slot free /
  slot busy same path / slot busy other path) and asserts the loader is up
  exactly when a read is in flight. Written because three review rounds
  each found a different unenumerated state of this machinery; the fourth
  is enumerated rather than waited for.
- **The loader frame** — an empty canvas reads as "no changes", the one
  answer this overlay must not give by accident.

## Screenshots / TUI captures

No terminal capture: the session that produced this is non-interactive,
so the overlay was driven through `TestBackend` rather than a real TTY.
The layout is pinned by the render tests above (title on the top rule,
padded tree body, counts right-aligned on the bottom rule, centred `j/k
scroll  g/G top/bottom  Esc close` hint line). **A visual pass on a real
terminal is still worth doing before merge.**

## Checklist

- [ ] README updated if CLI surface changed (n/a: TUI-only)
- [ ] `examples/gwm.toml.example` updated if config schema changed (n/a:
  no new config key; the new `[tui.keys.modal.working_tree]` context is
  documented in the config reference like the other modal contexts)

## Linked issues / docs

- Issues: #592, #613
- Twin still open: #593 (the commit listing full size). Deliberately not
  factored into a shared shell here: one implementation, no abstraction
  invented for a second caller that does not exist yet. It will hit the
  same sidebar-cache trap.

## Notes for reviewers

Three things that are pre-existing scope rather than regressions:

- The overlay is **bordered** rather than following `[tui] layout`. True
  of every modal today; #594's scope.
- The listing keeps the pane's `WT_TREE_MAX_FILES = 500` cap and its
  `… N more` row. Same tree the pane builds, at a bigger size; the cap is
  not raised.
- The **config-reference correction** reaches past this feature. The
  paragraph told users to pick the array form *or* the table form for
  `create` / `help` / `command_logs` / `link`, on the grounds that TOML
  forbids the same key twice. That describes the world before #219: the
  two live at different paths (`tui.keys.<name>` vs
  `tui.keys.modal.<name>`) and `config.rs` skips the `modal` namespace
  precisely so both survive the merge. Caught by Copilot when this PR
  extended the stale list with `working_tree`; corrected in both
  languages rather than extended.

**Review sources.** Copilot went three rounds:

- `2004813`, five findings: the #613 dispatch bug, the stale config
  paragraph EN + FR, the placement-blind count assertion, a test comment
  that described the opposite of the code.
- `7551808`, four: the inline `git status`, the missing multi-chord
  fallback, a rustdoc adopted by the function inserted above it, a test
  comment describing the ordering #613 had just replaced.
- `49cf7ae`, one: coalescing across worktrees leaving the loader up.

All addressed. Each round found a different unenumerated state of the new
async / dispatch machinery, so the last commit writes that state table
down as a test instead of inviting a fourth round.

Codex cloud and Codex CLI were both quota-blocked. CodeRabbit CLI ran
clean on the branch diff (0 findings across 17 files) before rate-limiting
on the re-run.

**CI note.** `pull_request: synchronize` has not fired on this PR since it
was opened: the only event-driven `ci` run is on `2004813`. Later heads
were validated by `workflow_dispatch` (`49cf7ae` green across all ten
jobs, `1f4b38e` running). Worth forcing a real synchronize before merge
if the branch protection wants the check on the head commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`5` opens the Working Tree listing at full size**
+  ([#592](https://github.com/kbrdn1/gwm-cli/issues/592)). The sidebar's
+  Working Tree pane is one block among five in a column that is a fraction of
+  the screen, so a worktree with more than a handful of changed files could
+  only be read two rows at a time through `J` / `K`. `5` now opens the same
+  file-explorer tree as a full-size overlay: same icons, same per-category
+  colours, the same change counts on the bottom rule, scrolled with
+  `j` / `k`, `g` / `G`, closed with `Esc` / `q` (or `5` again), and
+  rebindable under `[tui.keys.modal.working_tree]`.
+
+  The listing is read when the overlay opens rather than taken from the
+  sidebar's cache, so it does not go blank in the two states where that cache
+  is never built: sidebar hidden, or the Details panel showing stashes.
 - **`o` on the agents overlay resumes the session in the multiplexer**
   ([#591](https://github.com/kbrdn1/gwm-cli/issues/591)). The overlay told you
   which agent was working where and then left you to get there by hand. `a`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,14 +12,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- **`5` opens the Working Tree listing at full size**
+- **`W` opens the Working Tree listing at full size**
   ([#592](https://github.com/kbrdn1/gwm-cli/issues/592)). The sidebar's
   Working Tree pane is one block among five in a column that is a fraction of
   the screen, so a worktree with more than a handful of changed files could
-  only be read two rows at a time through `J` / `K`. `5` now opens the same
+  only be read two rows at a time through `J` / `K`. `W` now opens the same
   file-explorer tree as a full-size overlay: same icons, same per-category
   colours, the same change counts on the bottom rule, scrolled with
-  `j` / `k`, `g` / `G`, closed with `Esc` / `q` (or `5` again, whatever `5`
+  `j` / `k`, `g` / `G`, closed with `Esc` / `q` (or `W` again, whatever `W`
   gets rebound to, see #613 below), and rebindable under
   `[tui.keys.modal.working_tree]`.
 
@@ -310,7 +310,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - **An overlay's toggle key closes it whatever it is bound to**
-  ([#613](https://github.com/kbrdn1/gwm-cli/issues/613)). `3`, `4` and `5`
+  ([#613](https://github.com/kbrdn1/gwm-cli/issues/613)). `3`, `4` and `W`
   each close the overlay they open, but the guard doing it asked
   `key_matches_action`, which reads a single stroke and only ran after the
   modal verbs had their turn. Two silent holes: a multi-stroke binding
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -254,6 +254,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   then scrolled it. The toggle now resolves first, against that one action
   rather than the whole keymap, and it accumulates its chord, so a prefix
   stroke is consumed instead of firing a scroll verb on the way through.
+
+  Each of the three overlays routes its keys through an `App` method now
+  (the shape the create overlay has had since #217), because the ordering
+  is the fix and a `match` in the run loop cannot be tested. `d` still
+  cannot reach the delete confirm from behind an overlay: the toggle
+  resolves against its one action, not the whole keymap.
 - **A compact pane's header says where you are, and its name no longer
   dims when it is not**
   ([#605](https://github.com/kbrdn1/gwm-cli/issues/605)). In the default
@@ -297,12 +303,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   Bordered mode is otherwise untouched: there the accent still paints the
   four rules and the title inside the top one.
-
-  Each of the three overlays routes its keys through an `App` method now
-  (the shape the create overlay has had since #217), because the ordering
-  is the fix and a `match` in the run loop cannot be tested. `d` still
-  cannot reach the delete confirm from behind an overlay: the toggle
-  resolves against its one action, not the whole keymap.
 
 - **A linked row with nothing fetched yet is white, not green and purple**
   ([#596](https://github.com/kbrdn1/gwm-cli/issues/596)). The table's `I/P`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   only be read two rows at a time through `J` / `K`. `5` now opens the same
   file-explorer tree as a full-size overlay: same icons, same per-category
   colours, the same change counts on the bottom rule, scrolled with
-  `j` / `k`, `g` / `G`, closed with `Esc` / `q` (or `5` again), and
-  rebindable under `[tui.keys.modal.working_tree]`.
+  `j` / `k`, `g` / `G`, closed with `Esc` / `q` (or `5` again, whatever `5`
+  gets rebound to, see #613 below), and rebindable under
+  `[tui.keys.modal.working_tree]`.
 
   The listing is read when the overlay opens rather than taken from the
   sidebar's cache, so it does not go blank in the two states where that cache
@@ -240,6 +241,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **An overlay's toggle key closes it whatever it is bound to**
+  ([#613](https://github.com/kbrdn1/gwm-cli/issues/613)). `3`, `4` and `5`
+  each close the overlay they open, but the guard doing it asked
+  `key_matches_action`, which reads a single stroke and only ran after the
+  modal verbs had their turn. Two silent holes: a multi-stroke binding
+  (`working_tree = ["g w"]`) could open the overlay and never shut it, and a
+  binding the overlay's own context already claimed (`= ["j"]`) opened it and
+  then scrolled it. The toggle now resolves first, against that one action
+  rather than the whole keymap, and it accumulates its chord, so a prefix
+  stroke is consumed instead of firing a scroll verb on the way through.
 - **A compact pane's header says where you are, and its name no longer
   dims when it is not**
   ([#605](https://github.com/kbrdn1/gwm-cli/issues/605)). In the default
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -295,6 +295,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Bordered mode is otherwise untouched: there the accent still paints the
   four rules and the title inside the top one.
 
+  Each of the three overlays routes its keys through an `App` method now
+  (the shape the create overlay has had since #217), because the ordering
+  is the fix and a `match` in the run loop cannot be tested. `d` still
+  cannot reach the delete confirm from behind an overlay: the toggle
+  resolves against its one action, not the whole keymap.
+
 - **A linked row with nothing fetched yet is white, not green and purple**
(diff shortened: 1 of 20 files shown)


kbrdn1:
@codex review

kbrdn1:
Rebased onto `dev` (merge commit `0a89c59`) and aligned with the commit listing #593 shipped, which landed while this was open.

**The merge.** #593 is the same overlay built twice, so every additive list both touch conflicted head on: `View`, `Action`, `KeyContext`, `TaskKind`, `TaskMsg`, `palette_entries`, the help rows, `sizing_matrix`, and the four docs tables. All resolved as the union. The one deletion is `resolve_contextual_action`, which #593 removed on purpose once `c` meant the same in both panes.

**`5` is now `W`.** A letter rather than the next number: `3` and `4` open gwm's own panels (the transcript, the settings), while this overlay and `c` / `C` open a pane's content. It is advertised in both pane footers as `tree`; `wt` was ruled out because it sat next to `wt scroll` and read as the same thing.

**Each row says how much the file changed.** `+120 -34` on the right, from one `git diff --numstat -z HEAD` in the same worker that reads the tree, so staged and unstaged are counted together and the rows never wait on a second read. #593 chains its stats because a `git log --raw --numstat` over 300 commits takes seconds; one diff against `HEAD` is quicker than the `git status` in front of it, so splitting would buy a flicker and cost a second payload identity to match on.

The column rides its own rect, right-aligned, and is dropped whole on a terminal too narrow to keep it beside a readable file name, so the name is never what goes. `commits_meta_pick` became `meta_pick(body_w, tiers, floor)`, the width policy both listings share.

One entry per row, **including the rows with nothing to count** (a directory, an untracked file, a binary file). That is the load-bearing invariant: skipping their slot slides every count below onto the wrong file, which a reader cannot spot. Proven red by deleting the `paths.push(None)` in the `Dir` arm, which fails the length assert 2 vs 3 rather than something incidental.

**`D` / `U`** page by half a screen, against the viewport the renderer publishes rather than half the content.

### Two things worth flagging

**This merge commit is not only a merge.** The alignment touches the same lines of `ui.rs` and `tui_app_tests.rs` as the conflict resolution, so separating them after the fact would have meant hunk surgery on a green tree. The commit body says so.

**It touches #593's footer.** `commits_footer_hints` gains the `D/U half page` entry its own help overlay already documented. One line, outside this issue's strict scope; say the word and it comes out.

### Not done, deliberately

No load-more. `WT_TREE_MAX_FILES = 500` exists for #300's untracked-directory explosion, and paging a pathology is not the same feature as paging history: the `… N more` row already tells the truth.

### Verification

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and the full `cargo test` (104 suites) all green locally. Visual validation still pending.

Issue #453, opened 2026-07-24
[Feature]: complete the help overlay with every modal context's keys


kbrdn1:
### Problem

The Keybindings overlay (`?`) documents the global actions and a handful of modal verbs (Create Form, Delete Worktree, the Issue / PR pickers), but the TUI now carries 14 modal key contexts and most of them are absent from the overlay. Their keys are only discoverable through `gwm tui keys`, the footer hints, or the Settings Keys tab.

Missing sections today:

- the overlay's own navigation (`help` context: scroll, top/bottom, close)
- Detail / Agent Sessions (`detail`: select, attach, detach, pin)
- Command Logs (`command_logs`)
- Settings panel and its inline editor (`config`, `config.edit`)
- Report (`report`)
- Command Palette (`palette`)
- Exec Picker (`exec`) and Clean (`clean`)
- the PTY overlay escape hatches

The completeness guard from #334 (`help_overlay_documents_every_action`) only pins `Action::all()`, so modal verbs can keep drifting out of the overlay without failing CI.

The same drift applies to the **statusline which-key hints**: `HintContext::hint_specs()` covers every context variant, but each list is a static hand-curated selection that has not been revisited as features landed. The per-context picks, their grouping and their right-to-left truncation priority (#290) should be re-audited against the current action set (exec / clean / agent sessions / mux / macros / the new Working Tree scroll, etc.), so the footer advertises the verbs that actually matter in each context instead of the historical selection.

### Proposed solution

- One section per modal context in the help overlay, resolved live against `ModalKeymap`: rebinds show through, and a verb the user explicitly unbound is dropped rather than advertised as a phantom literal (same rule the existing Issue / PR rows already follow).
- Extend the completeness guard so every `ModalAction` (or at least every `KeyContext`) must surface in a help section, mirroring `help_overlay_documents_every_action`.
- Keep the section order stable: Global, List View, then modals in workflow order.
- Revise the statusline which-key hints per context (`hint_specs()`): re-audit each context's curated list against the current code, regroup by family, and re-rank the truncation order so the most useful verbs survive on narrow terminals. Pin the revised picks with the existing footer-order tests (`tui_footer_tests.rs`).

### Alternatives considered

- Pointing the overlay at `gwm tui keys` output: loses the curated grouping and the live rebind resolution the overlay already does for list-view actions.
- Documenting only the most-used modals: the drift problem stays; the guard is the point.
[...]

PR #614, opened 2026-08-26
✨ feat(tui): the commit listing, given the whole screen and paged

### LINKED ISSUE #593: [Feature]: a keybinding that opens the commit listing full size, with load-more

kbrdn1:
## Description

The sidebar's Commits pane is a fraction of a sidebar shared with four other blocks, and it stops at `RECENT_COMMITS_LIMIT` (300): reading further meant leaving gwm for lazygit. `6` now paints the same graph renderer on a 90% canvas, and `m` re-reads one page deeper up to 1500, so history is paged rather than capped.

Closes #593

## Type of change

## Changes

- `View::Commits` + `Action::Commits` (`commits`, default `6`) + `KeyContext::Commits` with `close` / `load_more` / `scroll_*` verbs under `[tui.keys.modal.commits]`, plus the palette entry and both help-overlay surfaces (list-view action and its own section).
- `CommitsModal` (`src/tui/state/commits.rs`): the snapshotted graph rows, the limit they were read at, the scroll cursor, the loader flag and the worktree path the payload must match. `COMMITS_PAGE` = the sidebar's own 300, `COMMITS_MAX` = 1500.
- The read runs on a `TaskKind::Commits` worker, never on the keypress. The revwalk sorts `TIME | TOPOLOGICAL`, so it traverses the whole reachable graph before yielding a row: **measured on this repo (2058 commits), asking for 300 costs the same as asking for all of them** (405ms vs 493ms, debug). The limit truncates the output, it bounds nothing about the latency. Paging is not cheap either, which is half the reason for the cap.
- The other half: `recent_commits_cached` is keyed on `(repo, tip, limit)`, holds 64 entries and evicts an arbitrary one when full, so unbounded paging would push other worktrees' sidebar entries out and force them to re-walk.
- `can_load_more` gates on the loader too: `loaded` still describes the previous page while a read is out, so a held `m` would raise the limit again and coalesce onto a read whose payload the drain then drops on the limit check, leaving the loader up forever. The drain checks both path and limit; a reopen differing in either frees the slot first.
- Title carries the row count and a trailing `+` while a deeper page exists; the footer drops `load more` when the key would do nothing and says `loading` while a read is out.
- Rows are snapshotted at open rather than read from `SidebarState::cache` (only rebuilt while the sidebar is open and in commits mode, so empty in the two states where the overlay is most useful). The first page reads at the sidebar's limit, so a warm tip makes the worker a hash lookup.

## Tests

- [x] `cargo test` passes locally (104 test binaries green, real exit code, `--no-fail-fast`)

Seven pins in `tests/tui_app_tests.rs` plus three render tests and the `sizing_matrix` entry. Every guard was falsified to prove its red:

- the width pin was faked (72/180 to 71/179) and both width guards named `commits`;
- dropping the cap from `can_load_more` reddened `paging_stops_at_the_cap`;
- removing the help section reddened the `#453` completeness guard (and surfaced that `#334` wanted a list-view `entry(Action::Commits, …)` too, which `cargo test`'s fail-fast had been hiding);
- removing the slot invalidation turned both reopen tests into a 10s timeout.

The two reopen tests stage the in-flight read by occupying the task slot rather than racing a real worker: raced, they pass with the invalidate removed, because on a one-commit fixture the thread finishes before the second open. That version was written first and thrown away.

Also ran the CI-like minimal-PATH pass (`PATH="$(dirname "$(command -v cargo)"):/usr/bin:/bin"`, after a normal-PATH `--no-run` since the macOS linker needs the full PATH to find `iconv`): green.

## Screenshots / TUI captures

None yet. **Visual validation is pending your local GO** (this session is not interactive), which is the usual gate for a TUI change here.

## Checklist

- [ ] README updated if CLI surface changed (n/a, no CLI surface change)
- [ ] `examples/gwm.toml.example` updated if config schema changed (n/a, `[tui.keys.modal.<context>]` takes any exposed context; the four docs pages carry the new one)

## Linked issues / docs

- Issue: #593
- Related PR: #612 (#592, the Working Tree modal)
- Docs: `docs/2.tui/1.keybindings.md`, `docs/4.configuration/1.gwm-toml.md` and both `docs/fr/` mirrors

## Notes for reviewers

**This takes `6`, not `5`.** #612 owns `5` for the Working Tree listing and is not merged yet, so `5` looks free on `dev` but is not. Branching this off `dev` rather than stacking on #612 was deliberate (#612 is currently `CONFLICTING` with uncommitted work in its worktree, so basing on it makes this unmergeable by construction). Whoever merges second resolves purely additive collisions: the `View` enum, `KeyContext::all()`, `palette_entries()`, `sizing_matrix()`, `help_rows`, the docs tables and the CHANGELOG. Until #612 lands, the numeric family reads `3`, `4`, `6` with a gap.

Two follow-ups deliberately left out of scope:

- The overlay does not share a shell with #592's. Two overlays is not an abstraction; if a third arrives, that is the moment to factor one.
- `Enter` on a row (copy the sha, open the commit on the forge) is in the issue as an explicit non-requirement for the first cut.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,11 +72,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the overlay works with the sidebar hidden or showing stashes. Scroll is
   `j`/`k`, `g`/`G`, all rebindable under `[tui.keys.modal.commits]`.
 
-  Each row carries the author and the commit age on its right, coloured by
-  freshness, which is what the hash / initials / subject columns do not say.
-  The column is dropped when the subject cannot spare the room: it narrows
-  to the age alone first, then disappears, so a narrow terminal keeps a
-  readable subject instead of buying a column with it.
+  Each row carries, on its right, what the hash / initials / subject columns
+  do not say: the author, what the commit changed (`3~ 1+ 2- +120 -34`, in
+  the Working Tree pane's colours, empty categories omitted) and how long
+  ago it landed. Three tiers, picked on what the **subject** can spare
+  rather than on the terminal width, since the graph is as wide as the
+  branch topology makes it: `author · counts · age`, `counts · age`, the age
+  alone, nothing.
+
+  The counts arrive from a second, chained read, so the log is on screen
+  immediately and the column grows about a second later (up to three on the
+  deepest page). One `git log --raw --numstat` over the rows already shown
+  costs about a second where a `diff_tree_to_tree` per commit costs
+  thirty-three, measured. `--diff-merges=first-parent` is load-bearing:
+  without it `git log` says nothing at all about a merge, and this project
+  merges rather than squashes.
+
+  `D` / `U` scroll half a screen, matching the rich PR view.
 
 - **Two settings for what a mux spawn opens, and where**
   ([#589](https://github.com/kbrdn1/gwm-cli/issues/589),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,10 +54,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the status bar, since resuming it in a second pane while it runs elsewhere
   may fork or refuse depending on the tool.
 
-- **`6` opens the commit listing full size, with load-more**
+- **`c` opens the commit listing full size, with load-more**
   ([#593](https://github.com/kbrdn1/gwm-cli/issues/593)). The sidebar's
   Commits pane is a fraction of a sidebar shared with four other blocks, and
-  it stops at 300 commits: seeing further meant leaving gwm for lazygit. `6`
+  it stops at 300 commits: seeing further meant leaving gwm for lazygit. `c`
   now paints the same graph on the whole canvas, and `m` re-reads one page
   deeper, up to 1500 commits, so history is paged rather than capped. The
   title carries the row count and a trailing `+` while a deeper page exists;
@@ -71,6 +71,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   which is only rebuilt while the sidebar is open and in `commits` mode, so
   the overlay works with the sidebar hidden or showing stashes. Scroll is
   `j`/`k`, `g`/`G`, all rebindable under `[tui.keys.modal.commits]`.
+
+  Each row carries the author and the commit age on its right, coloured by
+  freshness, which is what the hash / initials / subject columns do not say.
+  The column is dropped when the subject cannot spare the room: it narrows
+  to the age alone first, then disappears, so a narrow terminal keeps a
+  readable subject instead of buying a column with it.
 
 - **Two settings for what a mux spawn opens, and where**
   ([#589](https://github.com/kbrdn1/gwm-cli/issues/589),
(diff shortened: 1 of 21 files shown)


kbrdn1:
## Codex review loop: stopped on stagnation, not on a clean pass

Four passes, `P0P1P2_COUNT=2` at the end. This is **not** a clean verdict and is not being reported as one.

|:--|:--|:--|:--|:--|:--|
| 1 | 2 | 0 | 0 | 1 | stranded loader when the selection drifts under the overlay — business code |
| 2 | 2 | 0 | 0 | 1 | inert `m` still advertised — on the line pass 1 wrote |
| 3 | 3 | 0 | 0 | 2 | tip missing from a read's identity, rows counted as commits — business code |
| 4 | 4 | 0 | 0 | 2 | both findings on the `head` field pass 3 added (`git blame` → `823338b`) |

**The turn:** pass 4. Both of its findings point at lines from my own pass-3 fix, and neither survives contact with the code.

**Fixed (4 commits, each with a red-verified guard):**

- `fed9e0b` — `load_more_commits` read `selected()` while the drain matches `commits.path`. An auto-refresh moving the selection under an open overlay left the loader up permanently. Test drifts the selection and times out at 10s without the fix.
- `3539853` — the footer offered `m` after the worktree left the list. `App::commits_can_load_more` is now the single predicate the renderer and the handler both read.
- `823338b` — (a) the tip joins path and limit in what identifies a read: close mid-read, land a commit, reopen, and the request rode a worker holding the old `head`. (b) `recent_commits_lines` paints one sentinel row for an unborn HEAD, so counting rows titled the overlay `Commits (1)` over a repo with none; `recent_commits_listing` carries the real count.
- `30eed12` — the rationale below, in the code.

**Declined, with the evidence:**

- *"Resolve HEAD when the overlay opens"* — `WorktreeInfo.head` is the snapshot `worktree::list` takes at each refresh, and that is the documented contract of `recent_commits_cached`. The sidebar's Commits pane hands **the same struct to the same memo** (`src/tui/ui.rs:1632`). Resolving HEAD here would make a full-size view of a pane disagree with the pane it magnifies. The staleness window is one auto-refresh interval.
- *"Reject a payload computed on a tip the refresh has moved past"* — applied literally, this leaves `loading` true with nothing left to clear it: the stranded loader passes 1 and 2 already fixed twice. Making it safe means rejecting *and* re-reading, i.e. an overlay that re-walks on every tip change, which #593 does not ask for. Reopening after a refresh already picks up the new tip, and that path has a guard and a test.

**Convergence signals:** self-generated findings (pass 4 on pass 3's lines, blame-confirmed); severity flat at P2 with no P0/P1 in any pass and P3 rising; three conjunctions in the surviving scenario ("a commit lands *during* the ~400ms walk **and** the auto-refresh fires in that window **and** it updates `head` before the worker returns"); one finding refutable from the source. Decisive: **none of this code has ever been run.** The next useful information comes from opening the overlay, not from a fifth pass.

Validation still pending: `6` opened by hand, `m` on a repo with more than 300 commits, and how the loader reads at that size.

PR #615, opened 2026-08-28
✨ feat(tui): open a link in a terminal browser when a multiplexer is available

### LINKED ISSUE #590: [Feature]: open Issue/PR in a terminal browser when a multiplexer is available
## Problem

Every URL the TUI opens goes through one function, `tui/mod.rs:1578-1590`:

```rust
fn open_url(url: &str, app: &mut App) {
  let opener = if cfg!(target_os = "macos") { "open" } else if cfg!(target_os = "windows") { "explorer" } else { "xdg-open" };
  ...
}
```

Its callers are the browse-links menu (`B`, `Action::BrowseLinks`), the open-menu picks for Issue and PR, the rich PR/Issue view row opens, the CI checks overlay row open, and `Action::OpenDocs` (`.`). All of them leave the terminal for the system browser, which on a tiling setup means losing the workspace gwm is sitting in.

Inside a multiplexer that is avoidable: the page can open in a pane next to gwm, in a terminal browser.

## Proposed solution

A `[tui]` setting naming a terminal browser, used **only when a multiplexer is detected**, with the system opener as the fallback everywhere else:

```toml
[tui]
# command that renders a URL inside the terminal; {url} is substituted
terminal_browser = "w3m {url}"     # or lynx / carbonyl / browsh / claude's terminal-browser
```

Behaviour:

- multiplexer detected (`$TMUX` / `$ZELLIJ` / `$HERDR_ENV`) **and** `terminal_browser` set: spawn the command in a new mux pane (reusing the `mux_pane` builders and the pane-direction setting).
- otherwise: today's `open_url`, unchanged.

The gate is what the title asks for: a terminal browser with nowhere to put it is worse than the system browser, so no multiplexer means no change in behaviour.

One decision to make before implementing: whether this is a generic `{url}` command template (works with w3m, lynx, carbonyl, browsh) or a binding for one specific tool. The template is the same shape as `[tui.open] editor_cmd` / `shell_cmd`, so it fits the existing config vocabulary.

## Surface

config (.gwm.toml schema) + tui

## Backward compatibility

No breaking change: unset means today's behaviour on every platform.

## Additional context

- `open_url` is the single funnel, so the whole change is one function plus a config read. Every caller inherits it.
- The URL comes from the forge layer and is already built by `github.rs` / `gitlab.rs`; it is not user text, but the command template is, so it goes through the same sanitising as the other `[tui.open]` commands (see the branch-name injection advisory GHSA-fffq-vg6f-gxqm for why that matters here).
- Worth checking whether the terminal browser should get a pane or a full-screen PTY overlay (`View::Pty` already hosts lazygit and the review tool). A pane keeps gwm visible, which is the point.

kbrdn1:
## Description

Every URL the TUI opens went through one function and one OS opener, so a link always left the terminal. On a tiling setup that means losing the workspace gwm is sitting in. `[tui] terminal_browser` names a command that renders the page in the terminal instead, and the multiplexer is the gate: with nowhere to put a terminal browser, the system browser is still the better answer.

```toml
[tui]
terminal_browser = "w3m {url}"   # or lynx / carbonyl / browsh
```

Where a URL lands:

| Situation | Where it opens |
|:----------|:---------------|
| `terminal_browser` unset (default) | System browser, silently. Exactly the behaviour up to 1.9. |
| No multiplexer (`$TMUX` / `$ZELLIJ` / `$HERDR_ENV` unset) | System browser, reason on the status bar. |
| Browser not on `$PATH` | System browser, status bar names the binary. |
| tmux, or a zellij pane | A new pane or tab running the browser beside gwm. |
| herdr, a zellij tab, any `workspace` | PTY overlay, with the refusing backend named. |

Two design calls worth a reviewer's eye:

- **herdr gets the overlay, not the sequenced round trip #591 uses.** herdr takes no command in the argv that opens a container, and its way around that (open, wait for the new shell's prompt, then type the line) measured up to ~60s on a `direnv` + nix worktree. For a browser that is the wrong trade: a keypress has to show a page. The overlay still renders it in the terminal, it just covers gwm, which the status bar says out loud.
- **The template is tokenised before `{url}` is substituted**, the opposite order from `expand_agent_resume`. That one constrains the value because it cannot control where the user's template put the placeholder. A URL carries `: / ? # % = & +` by construction, so an allowlist wide enough for a real one would admit `&`. Splitting first makes `w3m {url}` and `w3m "{url}"` the same command and leaves the URL as one argv element gwm quotes itself. See GHSA-fffq-vg6f-gxqm for why the order matters.

Closes #590

## Type of change

- [ ] 🐛 Fix (bug fix)
- [ ] 📝 Docs
- [ ] ✅ Tests
- [ ] ⚡ Perf
- [ ] 🔧 Chore / CI / build

## Changes

- `[tui] terminal_browser` in the config schema, `"" == omitted` like `[tui.open] shell_cmd`.
- `expand_terminal_browser` / `is_browsable_url`: split first, substitute single pass into a token, append the URL when the template has no `{url}`.
- `plan_terminal_browser` + `BrowserPlan`: mux pane, PTY overlay, or OS opener, with the reason carried whenever a browser was configured and could not be honoured. env values and the `$PATH` probe are parameters, so every branch is drivable in a test without rewriting a process-global `$TMUX` the clipboard path also reads.
- `open_url` routes through the plan; `PtyKind::Browser` so the overlay title says what it is.
- `SettingField::TerminalBrowser` under the Settings **TUI** tab, `Text` so the empty string (turning it back off) is reachable.
- Docs: `.gwm.toml` reference EN + FR, `examples/gwm.toml.example`, CHANGELOG under `[Unreleased] > Added`.

## Tests

- [x] `cargo test` passes locally (104 suites, 0 failures)

13 new tests. The three that carry the load:

- `the_planned_browser_line_reaches_a_real_shell_as_one_argument` runs the planned line through `/bin/sh` with `printf '%s\n'` as the browser, so the quoting claim is checked against a shell rather than against a `Vec`. A hostile URL comes back as exactly one line, verbatim, and nothing in it executes.
- `terminal_browser_expansion_keeps_the_url_as_one_argument` pins that `w3m {url}` and `w3m "{url}"` are indistinguishable.
- `terminal_browser_that_is_not_installed_is_caught_before_the_spawn` pins the *reason*, not just the `System` variant, since three different branches produce that variant.

Each was checked red by mutation, not just observed green: substituting before the split fails the first two, dropping the `$PATH` probe fails the third.

## Screenshots / TUI captures

None. The visible surface is a status-bar line and a pane that a capture cannot show without a live multiplexer. **Not validated by hand yet** on a real tmux / herdr session, only under test.

## Checklist

## Linked issues / docs

- Issue: #590
- Builds on: #588 (multiplexer detection), #589 / #608 (`mux_open_in`, `mux_pane_direction`), #591 (why herdr does not get a pane here)

## Notes for reviewers

The issue left one decision open: a generic `{url}` command template, or a binding for one specific tool. This takes the template, because it is the shape `[git_tui] command` and `[tui.agent_resume]` already use.

The overlay rung for herdr goes beyond a strict reading of the issue, which only contrasts a pane with the system browser. Without it the feature is inert for herdr users, and an opt-in that does nothing reads as broken.


kbrdn1:
Two additions after a self-review, both pushed:

- **`gwm doctor` probes the configured browser** (`59fec33`). `[tui] terminal_browser` is a third configurable binary next to `[git_tui]` and `[review]`, and the PATH check could not see it. Opt-in like `[review]`, warning never failure, `{url}` goes through `extract_launcher_binary` so the placeholder is never reported as a missing binary. #392 is the precedent for treating a config surface doctor ignores as a defect.
- **The `All` tab is pinned** (`59f7789`). `resolved_rows` derives its rows from the serialised `Config`, so nothing was checking that `tui.terminal_browser` survives the `Option` round trip when it *is* set, which is the half the panel's source attribution reads.

One thing I checked and want on the record, since it is where a new `PtyKind` variant usually goes wrong: the overlay's key forwarding is **not** gated on the kind. `View::Pty` forwards every key to the child unless `p.finished` is set, and `mark_finished` is only ever called from the `PtyKind::Exec` arm. So `Browser` behaves like `LazyGit` / `Terminal`: keys reach w3m, `Esc` is the emergency detach, and child exit closes the overlay. Nothing to change, but it is worth a second pair of eyes rather than a claim.

Still not validated by hand on a live tmux / herdr session.

PR #303, opened 2026-06-15
✨ feat(tui): multi-repo workspace mode (#36)

kbrdn1:
## Description

Multi-repo **workspace mode**: operate across every git repo one level below a
root directory instead of a single repo. Built in one PR (CLI + TUI) per the
issue's full scope, TDD throughout.

Closes #36

## Type of change

- [x] ✨ Feature (new functionality)

## Changes

- **`workspace` module** (pure): `discover` walks one level under a root,
  opening each direct-child dir as a repo (ignores non-dirs / non-repos / bare,
  sorted by name); `merge_worktrees` flattens every repo's `worktree::list`
  into one repo-tagged table; `autodetect` is the bare-`gwm` trigger.
- **CLI**: global `--workspace <DIR>` flag; `gwm list --workspace` prints the
  merged table with a leading `REPO` column (per-row `--detect-pr`, `--format
  names` → `<repo>/<name>`); empty root → `GwmError::EmptyWorkspace`.
- **`gwm create --repo <name>`** required in workspace mode to pick the target
  child repo (lists candidates when absent/unknown).
- **Auto-detect**: bare `gwm` in a repo-free dir that holds child repos prompts
  `Open <dir> as a workspace? [Y/n]`; declines silently without a tty (pipes/CI
  keep the old single-repo behaviour).
- **TUI workspace mode**: `App` gains a `WorkspaceState` (per-repo session meta
  + per-row repo map); the active repo (`repo`/`workdir`/`config`/`repo_name`)
  follows the selection via `sync_active_repo` (re-opens the selected row's repo
  before each draw), so every selection-driven action operates on the right
  repo with no per-handler routing. `refresh` re-lists every repo synchronously
  in workspace mode. The list table renders a `REPO` column. Keymap/theme stay
  session-level (resolved once from the first repo).

## Tests

- [x] `cargo test` passes locally (full suite: 67 binaries green, exit 0)
- [x] `cargo fmt --check` passes
- [x] `cargo clippy --all-targets -- -D warnings` passes
- [x] New tests added under `tests/` (TDD red→green per phase):
  - `tests/workspace_tests.rs` — discover / merge / autodetect (pure + git2 tempdirs)
  - `tests/cli_workspace_tests.rs` — `list`/`create`/autodetect-decline E2E
  - `tests/tui_workspace_tests.rs` — merged-list construction, swap-on-navigation, REPO-column render
- New tests also re-validated under a stripped CI-like `PATH` (env-independent).

## Notes for reviewers

[...]

Issue #255, opened 2026-06-07
[Task]: migrate the GitHub fetch onto the async-task spine (#231 follow-up)


kbrdn1:
Follow-up to #231 (merged in #254). The generic `state::async_task::TaskRunner` spine (coalescing + generation late-drop) now backs the off-thread worktree refresh, and the statusbar loader render is already unified. The #217 GitHub fetch still runs on its own `github_tx`/`github_rx` channel with its own `inflight` HashSet dedupe.

**Goal:** move the GitHub fetch threading onto the shared spine so there is one off-thread mechanism, not two.

**Notes**
- The GitHub layer is also a per-`(target, number)` *cache*; the spine deliberately dropped the cache. Reconcile: either keep the cache as a github-specific concern layered over the spine, or model cache keys as part of `TaskKind`.
- Keep the deterministic test pattern (`*_result_sender()` + drain, no real OS thread).
- Render is already shared (`is_github_loading() || is_task_loading()`), so this is threading-only and should be behaviour-preserving.

kbrdn1:
## Codex adversarial-review finding (surfaced during #232 / PR #260)

A Codex adversarial review run while landing the Configuration panel
(#232) compared the branch against `main` and flagged a **pre-existing**
race in the current (pre-spine) GitHub fetch — exactly the gap migrating
onto the `TaskRunner` async-task spine would close. Capturing it here so
the implementer has the repro + a regression test ready.

**Verdict:** needs-attention (medium)

### Stale same-key GitHub fetches can win a retry race
`src/tui/state/github_fetch.rs:286-305`

`complete_issue` / `complete_pr` validate a returned worker result only by
removing `FetchKey::{Issue,Pr}(number)` from `inflight`. The worker
message carries **no generation**, so:

1. start a fetch for issue/PR `N`,
2. invalidate it (navigate away, or press refresh again),
3. start a new fetch for the same `N`,
4. the **old** worker returns first → this code sees the new inflight slot,
   accepts the stale result, clears loading, and the newer worker's result
   is later dropped.

**Impact:** the UI can report GitHub status / checks as refreshed while
actually displaying older data — hard to spot under slow `gh` / network
conditions.

### Recommendation (this is what the spine migration buys us)
- Add a monotonic generation / request id per `FetchKey`, include it in
  `GithubFetchMsg`, and require `complete_*` to match **both key and
  generation** before applying — i.e. mirror the generation guard
[...]

Issue #138, opened 2026-05-22
[Bug]: GitHubFetch cache identity ignores FetchKey number + complete_* races with invalidate


kbrdn1:
## Problem

Two correctness bugs flagged by Copilot review on PR #137 (the #128 extraction). Both involve the `GitHubFetch` sub-struct that PR introduced:

### Bug 1: `is_cached` cache identity collision

\`GitHubFetch::is_cached\` (\`src/tui/state/github_fetch.rs:261\`) treats any terminal \`issue_state\` / \`pr_state\` as a cache hit for **all** keys of that target — the lookup ignores the \`FetchKey\` number field. After:

\`\`\`rust
fetch.request(FetchKey::Issue(42));
fetch.complete_issue(42, Ok(...));
\`\`\`

a subsequent \`fetch.request(FetchKey::Issue(43))\` incorrectly returns \`HitCache\` — even though Issue 43 was never fetched. The dedupe contract promised by #128 was "(target, number) is the dedupe identity — Issue(42) and Pr(42) never collide" — that holds for the inflight path but breaks on the cache path.

### Bug 2: \`complete_*\` races with \`invalidate()\`

\`complete_issue\`/\`complete_pr\` (\`src/tui/state/github_fetch.rs:225\`) unconditionally stamp the result, even if the inflight slot was cleared by an intervening \`invalidate()\`. The race:

1. User visits worktree A → \`request(Issue(42))\` → Spawn → shell-out kicks off
2. User navigates to worktree B → \`invalidate()\` clears the inflight slot
3. The Issue(42) shell-out completes → \`complete_issue(42, Ok(...))\` runs → stamps stale \`Loaded(IssueStatus { ... worktree A's issue ... })\` into the now-active worktree-B state.

The cache is now incorrect AND \`invalidate()\` is no longer authoritative.

## Proposed fix

Single shape change: replace the separate \`issue_state: GitHubFetchState<IssueStatus>\` + \`pr_state: GitHubFetchState<PrStatus>\` fields with a per-key cache:

\`\`\`rust
pub struct GitHubFetch {
  pub link: BranchLink,
  pub link_slug: Option<String>,
  issue_cache: HashMap<u64, GitHubFetchState<IssueStatus>>,
  pr_cache: HashMap<u64, GitHubFetchState<PrStatus>>,
  inflight: HashSet<FetchKey>,
}
\`\`\`

[...]

Issue #528, opened 2026-08-04
[Feature]: inline review comments in the rich PR view (GraphQL)


kbrdn1:
## Problem

The rich PR / issue view landed with #420 shows the description, the metadata,
the CI rollup, the submitted reviews and the **conversation** comments. What it
cannot show is the part a review is usually about: the comments anchored to a
diff hunk.

That is not an oversight in the fetch, it is a hard boundary of the REST-ish
surface. `gh pr view --json comments` returns conversation comments only.
Review threads, their per-comment `path` / `diffHunk` / `line` / `startLine`,
and the reply chains inside them are reachable **through GraphQL alone**. This
is the one finding `snacks.nvim` / `snacks.gh` paid for up front, which is why
it ships a hand-written query (`api.lua:448-500`) next to its `gh --json`
calls rather than one uniform transport.

#420 shipped the other two thirds first on purpose: they cost one widened
request against a call gwm already makes, while this one is a second transport.

## Proposed solution

A `fetch_pr_threads` over `gh api graphql`, using the reference query as the
starting point:

- `reviewThreads -> comments { path, diffHunk, line, startLine, subjectType }`
- reply chains kept, so a thread renders as a thread and not as N loose rows

Plus a normalisation step on the `serde` side, the equivalent of snacks'
`clean_graphql`: GraphQL wraps everything in `nodes` and reaction groups, and
flattening those into the shape the REST parser already produces is what lets
**one** renderer serve both transports. Without it the rich view grows a second
row-building path for the same content.

Rendering slots into the existing `rich_view` builders as a `threads` section:
`path:line` as the header row, the diff hunk as context, the comments under it.
The caps and the wrapping are already there.

## Alternatives considered

- **Leave it out.** Defensible: `Enter` on a comment row already opens the
  thread in the browser, so nothing is unreachable, only slower. The counter is
  that "read the review without leaving the TUI" is exactly the promise of the
  view, and a review whose actionable half lives in the browser only half keeps
  it.
- **Poll the REST `/pulls/:n/comments` endpoint** instead of GraphQL. It does
  return inline comments, but flat: reconstructing which thread a reply belongs
  to means a second pass over `in_reply_to_id`, and it still misses the thread
  resolution state. GraphQL returns the shape already assembled.

## Surface

TUI (the rich PR view), and the GitHub backend. No CLI surface.

## Backward compatibility

Additive. A repo whose `gh` cannot run the query, or a GitLab remote, keeps the
[...]

PR #618, opened 2026-08-28
🐛 fix(tui): the selected worktree keeps the GitHub context fetched for it

### LINKED ISSUE #597: [Fix]: the selected worktree's context is not loaded, so C/c report nothing to show
## Problem

Standing on a worktree in the list, several context-dependent verbs report that they have nothing to show, and a manual refresh does not fix it. The clearest case is the CI checks overlay (`C`, and `c` from the status pane):

```rust
// tui/app.rs:3176-3187, enter_ci_checks
let checks = match self.pr_fetch_state() {
  GitHubFetchState::Loaded(pr) if !pr.checks.is_empty() => pr.checks.clone(),
  _ => {
    self.status = format!("no CI checks to show: link a PR and fetch ({key}) first");
    return;
  }
};
```

The two refresh keys do different things and neither fills that state reliably:

- `f` (`Action::Refresh`) refreshes the **worktree list** (paths, statuses, ages). It does not touch the GitHub layer.
- `F` (`Action::FetchGithub`) runs `refresh_github_status` (`tui/app.rs:6149`), which re-resolves the PR for the selected branch and loads its state, checks included.

So the CI overlay only works after an explicit `F` on the right row, and the same applies to anything else reading `pr_fetch_state`: the rich PR/Issue view (`I`), the CI state in the status pane, and the `I/P` pastille colours on the non-selected rows.

## Proposed solution

Load the selected worktree's context completely, without asking the user which refresh key means which half:

- fetch the GitHub context for the selected row when the selection changes (debounced, on the async-task spine so the frame never blocks), or
- make `f` fill both halves rather than the list alone, or
- keep the split but stop reporting "not available" for a state that is merely "not fetched yet": say "fetching" and kick one off.

The third option is the smallest and may be enough on its own: the user reads the message as a missing link, when the real state is "nobody asked yet".

Worth pinning as part of this: which verbs depend on a fetched state, and what each one shows in the unfetched case. From a first read: `C` / `c` (CI checks), `I` (rich view), the status pane's CI line, and the pastille placeholder colours.

## Surface

tui (behaviour)

## Backward compatibility

No breaking change in the config sense. An automatic fetch on selection changes how often gwm calls the forge CLI, which is a real cost on a big list: it needs the coalescing and late-result-drop the async spine already provides (#231), and probably a cache TTL.

## Additional context

- `refresh_github_status` is already careful about staleness: a failed probe keeps the persisted detection rather than blanking the pane (#284), and the per-key generation guard drops late results (#231).
- Workspace mode has its own refusal path (`workspace_active_stale`), which must keep refusing rather than fetching against the previously active repo (#304).
- Whatever lands, the acceptance test is behavioural: open the TUI on a worktree with an open PR, press nothing, and check what `C` does.

kbrdn1:
## Description

Standing on any row but the one the TUI opened on, `C` / `c` reported `no CI
checks to show: link a PR and fetch (F) first` for a worktree whose PR was
linked and whose checks had already been fetched, and `I` refused for the same
reason. `f` did not help: it refreshes the worktree list, not the GitHub layer.

gwm was throwing away its own prefetch. It fetches every linked issue and PR at
startup and on every relist, but `refresh_link` (which runs on every selection
change) flushed the whole result cache and dropped every in-flight `gh` worker
with it, so the prefetch died on the first `j`. That flush was a leftover of the
pre-#138 single-slot cache: results are keyed by `(side, number)` now, so a row
reads its own number and a number it never fetched reads `Idle`. It cannot serve
one row's status for another, which is the only thing the flush was defending.

Closes #597

## Type of change

- [x] 🐛 Fix (bug fix)

## Changes

- `App::refresh_link` keeps the fetched statuses and the in-flight workers.
  `reread_link` already reports the one case where a cached number means
  something else (an origin moving between two forge instances), so the flush
  and the spine generation-bump it must move with fire there and nowhere else.
  `GitHubFetch::refresh_link`, the wholesale-clearing wrapper, had one caller
  and is removed.
- The relist keeps that flush, and it moves ahead of every early return in
  `refresh_linked_github_statuses_for_worktrees` so the *relist* expires the
  statuses rather than the prefetch that happens to refill them. Workspace mode
  takes that early return by design (one slug cannot resolve every child repo's
  numbers, #303), so without the move it would have had no expiry at all once
  navigation stopped providing one.
- `forge_fetch_gap` answers, for one side of the link, why a context verb cannot
  run: a linked side never fetched is fetched now (same spawn helpers, same slug
  guard the bulk prefetch uses) and reported as in flight; one already in flight
  says the same without a second call; one whose probe failed reports what `gh`
  said, since `Error` is terminal in the cache and the "fetch first" hint pointed
  at a fetch that had already run. Only a row with nothing linked keeps the link
  hint.
- `enter_ci_checks` and `enter_rich_view` route through it. The CI overlay also
  splits out the fetched-but-empty rollup, which shared the link hint and so
  named a link and a fetch that were both already done.

## Tests

- [x] `cargo test` passes locally (full suite, exit 0)

Ten new tests, written red first. Each asserts its `Loaded` / unfetched
precondition **before** the action, so none can pass vacuously on a harness that
never had the state.

In `tests/tui_app_tests.rs`: three replace `refresh_link_invalidates_fetch_state`,
whose subject was the pre-#138 single-slot cache. A status survives a link
re-read, survives walking away and back (and does not leak onto the row in
between), and an in-flight worker survives the navigation that used to drop its
result. Six more cover the verbs: `C` and `I` starting the fetch nobody asked
for, reporting one in flight, reporting a failed probe, and the empty rollup.

In `tests/tui_workspace_tests.rs`: one pins the staleness bound on the path that
takes the prefetch's early return.

Two existing tests were re-aimed rather than deleted, because they were reading
the flush as if it were their subject: `a_link_refresh_drops_cached_threads_with_everything_else`
becomes `cached_threads_move_with_the_pr_they_hang_from` and now pins both halves
(kept on a re-read, dropped when the instance moves), and the filter test pins
re-resolution rather than invalidation. That one also stopped typing a single
`z`, which the tempdir's random name could match too, making it order-dependent.

## Screenshots / TUI captures

None: the change is in what the status bar says and what state survives a
keypress, both covered by the state tests above.

## Checklist

- [x] README updated if CLI surface changed (n/a: no CLI surface change)
- [x] `examples/gwm.toml.example` updated if config schema changed (n/a)

## Linked issues / docs

- Issue: #597
- Context: #138 (per-number cache), #255 (task spine), #303 (why the workspace prefetch is skipped), #458 (forge identity), #68 (the flush this removes)

## Notes for reviewers

Two things worth a second pair of eyes.

**Workspace ordering.** In workspace mode `on_navigation` runs `refresh_link`
*before* `sync_active_repo` swaps `self.repo` (`app.rs:1241`), so that first
re-read resolves the forge against the old repo. The swap then calls
`refresh_link` again, the identity moves, and the caches drop, so a preserved
cache cannot be served across repos. Both calls happen before the next draw
(`tui/mod.rs:314`), so the intermediate state never renders.

**Cost and freshness.** No new forge calls on navigation: nothing is fetched
that the bulk prefetch would not already have fetched, and the on-demand kick is
bounded by what the user actually opens. Staleness stays bounded by
`tui.auto_refresh_secs` (60 by default) in both modes, which is what the moved
flush buys. So no debounce and no TTL, which the issue floated for the
fetch-on-selection option this deliberately does not take.

Workspace mode gains the most: it skips the bulk prefetch by design, so before
this the context verbs there were fed by nothing at all.

Not validated by eye yet: this wants a local install and a walk through a list
with a linked PR before it merges.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -374,6 +374,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   already done. Only a row with nothing linked still gets the link hint.
   Workspace mode gains the most: it skips the bulk prefetch by design, so
   before this the verbs there were fed by nothing at all.
+
+  Freshness is unchanged: a relist still expires every fetched status, so
+  `tui.auto_refresh_secs` (60 by default) still bounds how stale one can be.
+  That expiry moved ahead of the bulk prefetch's early returns, which is what
+  gives workspace mode the same bound rather than none.
 - **An overlay's toggle key closes it whatever it is bound to**
   ([#613](https://github.com/kbrdn1/gwm-cli/issues/613)). `3`, `4` and `W`
   each close the overlay they open, but the guard doing it asked
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -346,6 +346,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **The selected worktree keeps the GitHub context that was fetched for it**
+  ([#597](https://github.com/kbrdn1/gwm-cli/issues/597)). Standing on any row
+  but the one the TUI opened on, `C` / `c` said "no CI checks to show: link a
+  PR and fetch (F) first" for a worktree whose PR was linked and whose checks
+  had already been fetched, and the rich Issue/PR view (`I`) refused for the
+  same reason. `f` did not help: it refreshes the worktree list, not the
+  GitHub layer, so only an `F` on that exact row filled the state back in.
+
+  gwm was throwing away its own prefetch. It fetches every linked issue and
+  PR at startup and on every relist, but the link re-read that runs on each
+  selection change flushed the whole result cache and dropped any in-flight
+  `gh` worker with it, so the prefetch died on the first `j`. That flush was
+  a leftover: the cache has been keyed by number since #138, so it cannot
+  serve one row's status for another, and a row that never fetched still
+  reads as unfetched. It is now dropped only when the origin actually moves
+  between two forge instances, which is the one case where a cached number
+  means something else.
+
+  The two verbs also stopped reading "nobody asked yet" as "nothing to show".
+  A linked PR that has never been fetched is now fetched on the spot, on the
+  same task spine, and reported as `fetching Pull request #61...`; one
+  already in flight says the same without starting a second call; one whose
+  probe failed shows what `gh` said instead of pointing at a fetch that had
+  already run; and a PR that is fetched with an empty rollup says its checks
+  have not been reported rather than naming a link and a fetch that are both
+  already done. Only a row with nothing linked still gets the link hint.
+  Workspace mode gains the most: it skips the bulk prefetch by design, so
+  before this the verbs there were fed by nothing at all.
 - **An overlay's toggle key closes it whatever it is bound to**
   ([#613](https://github.com/kbrdn1/gwm-cli/issues/613)). `3`, `4` and `W`
   each close the overlay they open, but the guard doing it asked
--- a/src/tui/app.rs
+++ b/src/tui/app.rs
@@ -7617,7 +7617,12 @@ impl App {
     // early return below, and it is what bounds staleness to
     // `tui.auto_refresh_secs`. Workspace mode needs it most, since it takes
     // the early return and refills per-selection instead.
-    self.invalidate_github();
+    //
+    // Settled entries only, and no spine generation-bump: a fetch in flight
+    // has no stale answer behind it to expire, so cancelling it would throw
+    // away work and expire nothing (Codex review on #618). The respawns
+    // below coalesce onto it through `TaskRunner::request`.
+    self.github.invalidate_settled();
 
     // Workspace mode (#36): this bulk prefetch resolves every merged row's
     // issue/PR against a single repo's slug (`self.github.link_slug`), which
--- a/src/tui/app.rs
+++ b/src/tui/app.rs
@@ -7181,22 +7181,31 @@ impl App {
 
   // ---- Issue/PR linking (issue #67) -------------------------------------
 
-  /// Re-read the link for the currently selected worktree's branch. Also
-  /// re-resolves the repo slug from the origin remote, and resets any
-  /// previously cached GitHub fetch state since it would refer to a
-  /// different (issue, pr) tuple now. Delegates to
-  /// [`GitHubFetch::refresh_link`] for the pure state mutation; the
-  /// branch resolution still lives here because it depends on
-  /// `App`'s `selected()` + `repo.head()` fallback.
+  /// Re-read the link for the currently selected worktree's branch, and
+  /// re-resolve the repo slug from the origin remote. Delegates to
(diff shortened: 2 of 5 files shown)


kbrdn1:
## Codex review — 3 passes, `P0P1P2_COUNT=0`

|:--|--:|--:|--:|--:|:--|
| 1 | 1 | 0 | 0 | 1 | workspace: le relist droppait un fetch en vol sans rien derrière |
| 2 | 2 | 0 | 0 | 2 | même classe, cause plus profonde + un bug préexistant |
| 3 | 0 | 0 | 0 | 0 | « Aucun défaut fonctionnel introduit » |

Le zéro de la passe 3 est corroboré : le rapport companion fait 158 caractères et
porte un verdict explicite, ce n'est pas un rapport vide compté comme zéro.

### Corrigé

**Passe 1 — un relist workspace annulait le fetch qu'il venait d'expirer** (`ae963fc` → `f9814a0`).
Le mode workspace saute le prefetch en masse par construction (#303), donc rien ne
remplaçait ce que l'expiration venait de jeter. `spawn_selected_github` redemande la
ligne sélectionnée : son lien et son slug appartiennent au repo *actif*, ce qui est
précisément la condition qui rend le passage en masse invalide là-bas. Il tourne après
le swap de repo, pas dans le tail partagé, où le lien est encore celui d'avant le relist.

**Passe 2 — un relist ne doit jamais expirer un fetch en vol** (`04400b9`).
Plus racine que le correctif précédent, qui traitait le symptôme. Ce qu'un relist
expire, c'est un résultat qui a cessé de faire autorité ; une entrée `Loading` n'en est
pas un — rien n'a encore été lu, il n'y a pas de réponse périmée derrière. Le flush la
jetait quand même, avec le bump de génération qui l'accompagne, donc le résultat du
worker était rejeté à l'atterrissage. Avec un `tui.auto_refresh_secs` plus court que la
latence du forge, chaque tick devançait le fetch : la ligne ne se remplissait jamais et
les sous-processus annulés continuaient de tourner. `invalidate_settled` garde les
entrées en vol et ne prend pas de bump ; les respawns derrière coalescent dessus via
`TaskRunner::request`.

Trois tests de plus dans `tests/tui_workspace_tests.rs`, écrits rouges d'abord, avec
leur précondition assertée avant l'action.

### Écarté

**Passe 2 — « relancez aussi les threads de la vue riche »** : réel, mais **préexistant
et hors scope**, suivi en #619.

Le relist vide `pr_threads_cache` et ne relance que issue + PR, donc les commentaires
inline disparaissent d'une vue riche ouverte quand la PR atterrit. Vérifié contre la
base plutôt que déduit :

```
$ git show origin/dev:src/tui/app.rs | grep -n 'invalidate_github\|spawn_github' …
7504:    self.invalidate_github();      # vidait déjà pr_threads_cache
7507:      if self.spawn_github_issue(n, &slug) {
7512:      if self.spawn_github_pr(n, &slug) {      # jamais PrThreads
```

Même comportement sur `dev`, en mono-repo comme en workspace, inchangé par cette PR.
`spawn_github_pr_threads` est délibérément hors du prefetch en masse depuis #528.

### Validation

`cargo test` complet exit 0, `cargo fmt --check` et `cargo clippy --all-targets -- -D warnings` verts.

